PR #3, opened 2026-02-03
Improve portability and fix minor bugs

CapriciousRebel:
## Summary
- Remove hardcoded user email/name from gitconfig for better portability across machines
- Fix perl regex in GPG key extraction (`print $&` instead of `print`)
- Make cargo env loading conditional to prevent errors when Rust isn't installed
- Simplify `link.sh` extract_and_link function and fix variable quoting (On macOs we have bash 3.2 installed by default, associative arrays are not supported on bash 4 and under)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


CapriciousRebel:
Hi @DJRHails, this project looks super cool!

I have added a few changes and minor fixes when trying to install it on my machine. Would aprreciate if you take a look at the PR 😄

PR #25, opened 2026-06-16
fix(slack): bot mode uses the canonical Slack host, ignoring SLACK_WORKSPACE

DJRHails:
## What

`_workspace_hosts()` checked `$SLACK_WORKSPACE` before the bot-token branch, so a bot (xoxb) invocation with `SLACK_WORKSPACE` set was routed to a per-workspace `<ws>.slack.com` host.

A bot token authenticates by the token itself (`Authorization: Bearer`), not by host — Slack routes the request to the token's own workspace regardless of which `<ws>.slack.com` it hits. The per-workspace host only matters for the personal cookie (xoxc) path, which probes hosts to find the one that accepts the cookie.

## Fix

Check the bot token first and return the canonical bare `slack.com` host unconditionally. Bot mode ignores `$SLACK_WORKSPACE`; the personal (xoxc) path still honours it.

## Tests

- New `test_bot_mode_ignores_slack_workspace_override` fails against the old ordering and passes with the fix.
- 4 regression tests added; `pytest test_richtext.py -q` → 35 passed. `ruff check` clean.

_(slack skill files are transcrypt-encrypted; the diff is ciphertext by design.)_


sage-saint:
## Review Summary

**Verdict: ✅ Approve.** Lane A (direct single-pass review — ~40 net hand-written lines, one focused function reorder + tests). No P1–P3 findings; nothing to fix or push.

The fix is correct and consistent with the surrounding code:

- `_workspace_hosts()` now checks `_bot_token()` **first** and returns `[""]` → bare `slack.com` (confirmed by `host = f"{ws}.slack.com" if ws else "slack.com"`). Bot mode no longer honours `$SLACK_WORKSPACE`, so the turnstile sentinel-swap (which matches only bare `slack.com`) fires.
- Making the `$SLACK_WORKSPACE` / `_DISCOVERED_WORKSPACE` branches unreachable in bot mode is **not** a regression: `_DISCOVERED_WORKSPACE` is set only in `_discover_token` (the xoxc path), which `_token` skips for bot tokens. The reorder also removes a latent inconsistency — previously `_token` would Bearer-auth with the bot token while `_workspace_hosts` routed to `<ws>.slack.com`; now they agree.
- **Bug-class sweep:** the only other `<ws>.slack.com` host construction (`_discover_token` / `_auth_test`, `auth_hosts`) runs solely in personal mode, so there's no sibling occurrence of this footgun in the changed surface.
- Tests follow falsify-first and cover all four branches (bot ±override, personal ±override).
- The PR's scoped-out follow-ups (drop `SLACK_WORKSPACE` from bot-mode loop invocations; optional `*.slack.com` turnstile rule) are reasonable and correctly deferred.

### Verification

- **Tests**: `pytest test_richtext.py -q` → **35 passed** (incl. the 4 new regression tests), run offline against uv's cached deps.
- **Lint**: `ruff check slack.py test_richtext.py` → **clean**.
- **Format**: `ruff format --check` flags drift, but it is **entirely pre-existing** (lines 163/178/386/398/2036 — untouched by this PR; repo ships no ruff config and doesn't enforce `ruff format`). The PR's changed region is format-clean. Out of scope.

_Note: these skill files are transcrypt-encrypted at rest, so the GitHub diff shows ciphertext (~2k± lines); the real change is the single `_workspace_hosts()` reorder above, reviewed against the decrypted plaintext._

_[via gantry](https://gantry.hails.info)_

PR #29, opened 2026-06-17
fix(python): write absolute init.templateDir so prek can use it

DJRHails:
## Problem
`modules/python/install.sh` wrote a **literal** `'~/.git-hooks'` into `~/.gitconfig.local`. git expands `~` in `init.templateDir`, but **prek** (the pre-commit drop-in our repos run) does **not** — so `prek init-templatedir` dies `os error 2` seeding the bogus path.

This was the root cause of a **fleet-wide gantry build outage**: every worker devcontainer rebuild failed at the prek step (Step 23, `devcontainer build failed (exit 1)`). Gantry has a workaround (DJRHails/gantry#287 rewrites the value in the build), but this fixes it at the source for every consumer.

## Fix
Write the **expanded absolute** `$HOME/.git-hooks`. `.gitconfig.local` is machine-local (regenerated per host by this script), so an absolute path is correct and works for both git and prek.

Verified against the real worker image: `init.templateDir` `~/.git-hooks` (prek RC 2) → `/home/dev/.git-hooks` (prek RC 0).

### DIFF
--- a/modules/python/install.sh
+++ b/modules/python/install.sh
@@ -33,9 +33,13 @@ if cmd_exists pre-commit; then
   # is machine-local, so it lives in gitconfig.local (see modules/git/gitconfig),
   # which the git module generates before this script runs.
   if [ -f "$HOME/.gitconfig.local" ]; then
-    # shellcheck disable=SC2088 # literal ~ is intentional: git expands it, keeps the entry host-portable
-    git config --file "$HOME/.gitconfig.local" init.templateDir '~/.git-hooks'
-    log::success "pre-commit templatedir initialised at ~/.git-hooks"
+    # Write the EXPANDED absolute path, not a literal '~/.git-hooks'. git expands ~
+    # in init.templateDir, but prek (the pre-commit drop-in our repos run) does NOT
+    # — a literal tilde makes `prek init-templatedir` die `os error 2` seeding the
+    # bogus path. .gitconfig.local is machine-local (regenerated per host here), so
+    # an absolute path is correct and keeps both git and prek happy.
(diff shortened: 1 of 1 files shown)


sage-saint:
## Review Summary

Post-merge review (PR was already merged into `main` at `0964e6ca` before this ran). Direct single-pass review — 1 file, +7/-3 shell change.

### Findings

| 1 | P4 | The `else`-branch `log::warning` (unchanged by this PR) still advises `set init.templateDir=~/.git-hooks` — the literal-tilde value this PR proved breaks prek. Misleading, but only fires when `.gitconfig.local` is missing (the git module generates it first). | Noted, not fixed — not worth a follow-up PR on a merged, working fix unless desired. |

No P1–P3 issues. The fix is correct and minimal:

- prek does **not** expand `~` in `init.templateDir`, so writing the expanded `$HOME/.git-hooks` is the right call; `.gitconfig.local` is machine-local, so the absolute path is the correct ownership boundary.
- Consistent with `modules/git/gitconfig:240-247`, which explicitly says machine-specific absolute `init.templateDir` paths belong in `.gitconfig.local`.
- No sibling occurrences of the bug class — this is the only `init.templateDir` writer in the repo.
- The now-unnecessary `SC2088` disable was correctly dropped on the fixed line.

### Verification

- **shellcheck -x `modules/python/install.sh`**: clean (RC 0)
- **Bug-class grep** (`templateDir` / `git-hooks` / literal-tilde paths): no other affected sites
- **Tests/build/format**: n/a — single shell-config edit; CI has no test for this path
- No fixes pushed (no blocking findings; PR already merged)

Verdict: approve

PR #30, opened 2026-06-22
feat(claude): add claude::resume; session-search emits it

DJRHails:
`claude::resume <id>` locates a session transcript across config-dir variants (`.claude` / `.claude-ant` / `.agents`), reads the real `cwd` from the transcript (the encoded project-dir name is lossy — both `/` and `.` collapse to `-`), `cd`s there, and launches through the profile that owns the session so `--resume` can find it.

`session-search` now prints `claude::resume <id>` instead of the bare `claude --resume <id>`, so a copy-pasted resume command lands you in the right directory automatically.

Verified: example session resolves + routes to the ant profile; a `.claude` session routes to the default profile; bad/missing args fail fast; `search.py` prints the new Resume line.

### DIFF
--- a/modules/claude/aliases.zsh
+++ b/modules/claude/aliases.zsh
@@ -61,6 +61,38 @@ claude::ant() {
     claude "$@"
 }
 
+# Resume a session by id, cd-ing into its original directory first. Claude can
+# only --resume a session from the cwd it was started in, and session-search
+# only prints the bare `claude --resume <id>`. This resolves both the directory
+# and the owning config profile so a copy-pasted id Just Works.
+claude::resume() {
+  emulate -L zsh
+  local id=$1
+  [[ -n $id ]] || { echo "claude::resume: usage: claude::resume <session-id>" >&2; return 1; }
+
+  # Search every config-dir variant (default + ant) and the legacy agents dir.
+  local file
+  file=$(command find "$HOME/.claude" "$HOME/.claude-ant" "$HOME/.agents" \
+           -type f -name "${id}.jsonl" -print 2>/dev/null | head -1)
+  [[ -n $file ]] || { echo "claude::resume: no session '$id' under ~/.claude*, ~/.agents" >&2; return 1; }
+
+  # The encoded project dir is lossy ('/' and '.' both collapse to '-'), so read
+  # the real working directory from the transcript's cwd field instead.
+  local dir
+  dir=$(command grep -m1 -o '"cwd":"[^"]*"' "$file" | cut -d'"' -f4)
+  [[ -n $dir && -d $dir ]] || { echo "claude::resume: cannot resolve cwd for '$id' ($file)" >&2; return 1; }
+
+  cd "$dir" || return 1
+
+  # Route through the profile whose config dir owns the session so --resume finds
+  # it (and the ant profile gets its env + self-healing links).
+  case ${file%%/projects/*} in
+    "$HOME/.claude-ant") claude::ant --resume "$id" ;;
+    "$HOME/.claude")     claude --resume "$id" ;;
+    *)                   CLAUDE_CONFIG_DIR=${file%%/projects/*} claude --resume "$id" ;;
+  esac
+}
+
 claude::local() {
   ANTHROPIC_BASE_URL=http://localhost:1234 \
   ANTHROPIC_AUTH_TOKEN=lmstudio \
--- a/modules/agents/skills/session-search/SKILL.md
+++ b/modules/agents/skills/session-search/SKILL.md
@@ -48,7 +48,7 @@ python3 ~/.claude/skills/session-search/scripts/search.py "deploy" --context 400
 
 For each matching session the script prints:
 
-- **Session ID** and resume command (`claude --resume <id>`)
+- **Session ID** and resume command (`claude::resume <id>` — cd's into the session's original directory and resumes via the owning config profile; defined in `modules/claude/aliases.zsh`)
 - **Date**, **project**, and the **original prompt**
 - Up to 3 **matching snippets** with role labels and line numbers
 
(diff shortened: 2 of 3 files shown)


sage-saint (commented):
Automated post-merge review of `claude::resume`. Two P3 findings inline; a follow-up PR with both fixes is linked in the summary. Solid, well-tested change overall.

modules/claude/aliases.zsh
```diff
@@ -61,6 +61,38 @@ claude::ant() {
...
+# and the owning config profile so a copy-pasted id Just Works.
+claude::resume() {
+  emulate -L zsh
+  local id=$1
+  [[ -n $id ]] || { echo "claude::resume: usage: claude::resume <session-id>" >&2; return 1; }
+
+  # Search every config-dir variant (default + ant) and the legacy agents dir.
+  local file
+  file=$(command find "$HOME/.claude" "$HOME/.claude-ant" "$HOME/.agents" \
+           -type f -name "${id}.jsonl" -print 2>/dev/null | head -1)
```
sage-saint, line 76: **P3 — performance / robustness:** this `find` walks the *entire* config trees on every resume — including `~/.claude/plugins` (which can hold a full node_modules tree), `todos/`, `shell-snapshots/`, and `statsig/` caches — when transcripts only ever live under `<config>/projects/<slug>/<id>.jsonl`. On a busy host that's a lot of wasted traversal per invocation.

Scoping to the `projects` roots fixes the speed and, as a bonus, removes the silent edge case in the `*)` branch below: if a stray `<id>.jsonl` were ever matched outside a `projects/` dir, `${file%%/projects/*}` returns the *whole path* unchanged and `CLAUDE_CONFIG_DIR` gets set to a file path. Restricting the search guarantees every match contains `/projects/`.

```zsh
file=$(command find "$HOME/.claude/projects" "$HOME/.claude-ant/projects" "$HOME/.agents/projects" \
         -type f -name "${id}.jsonl" -print 2>/dev/null | head -1)
```
(Missing roots just emit a suppressed stderr error, same as today.)

modules/claude/aliases.zsh
```diff
@@ -61,6 +61,38 @@ claude::ant() {
     claude "$@"
 }
 
+# Resume a session by id, cd-ing into its original directory first. Claude can
+# only --resume a session from the cwd it was started in, and session-search
+# only prints the bare `claude --resume <id>`. This resolves both the directory
```
sage-saint, line 66: **P3 — comment rot:** this line states "session-search **only prints the bare** `claude --resume <id>`" as the motivation — but this same PR updates `search.py` (`_format_location`) to emit `claude::resume <id>`. So the moment this merges, the justification describes a state that no longer exists. Suggest rewording to drop the stale claim, e.g.:

```zsh
# Resume a session by id, cd-ing into its original directory first. Claude can
# only --resume a session from the cwd it was started in; this resolves both the
# directory and the owning config profile from the id alone, so the command
# session-search prints Just Works on copy-paste.
```

sage-saint:
## Review Summary

Post-merge review of `claude::resume` + the `session-search` emit change. Direct single-pass review (34/−2 across 3 files — below the multi-agent threshold). The change is solid and clearly tested; only two minor P3 nits, both inline. **No correctness or security blockers.**

> ⚠️ This PR was already **merged** when the review ran, and the review identity (`sage-saint`) lacks push access to this repo, so I could not push the fixes to a branch. The patch is included below for one-command application.

### Findings

| F1 | P3 | `claude::resume`'s `find` walks the entire `~/.claude*` / `~/.agents` trees per call (incl. `plugins/` node_modules, `todos/`, `shell-snapshots/`, `statsig/`). Transcripts only live under `<config>/projects/`. Also: scoping removes the silent edge case where the `*)` routing branch's `${file%%/projects/*}` returns a full path as `CLAUDE_CONFIG_DIR`. | Fix prepared (patch below) — couldn't push |
| F2 | P3 | Header comment claims *"session-search only prints the bare `claude --resume <id>`"*, but this same PR updated `search.py` to emit `claude::resume`, making the stated motivation false at merge. | Fix prepared (patch below) — couldn't push |
| F3 | P4 | `cd "$dir"` persists in the caller's interactive shell after `claude` exits (it's a function, not a subshell). Likely intended (you land in the project dir), noted only. | No change |

### Apply the fix

```sh
git switch main && git pull && git switch -c djrhails/claude-resume-followup && git apply <<'PATCH'
diff --git a/modules/claude/aliases.zsh b/modules/claude/aliases.zsh
index 8aaa805f33bf..85286e694263 100644
--- a/modules/claude/aliases.zsh
+++ b/modules/claude/aliases.zsh
@@ -62,17 +62,20 @@ claude::ant() {
 }
 
 # Resume a session by id, cd-ing into its original directory first. Claude can
-# only --resume a session from the cwd it was started in, and session-search
-# only prints the bare `claude --resume <id>`. This resolves both the directory
-# and the owning config profile so a copy-pasted id Just Works.
+# only --resume a session from the cwd it was started in; this resolves both the
+# directory and the owning config profile from the id alone, so the command
+# session-search prints Just Works on copy-paste.
 claude::resume() {
   emulate -L zsh
   local id=$1
   [[ -n $id ]] || { echo "claude::resume: usage: claude::resume <session-id>" >&2; return 1; }
 
-  # Search every config-dir variant (default + ant) and the legacy agents dir.
+  # Search the projects tree of every config-dir variant (default + ant) and the
+  # legacy agents dir. Transcripts only ever live under <config>/projects/, so
+  # scoping here skips plugins/node_modules, todos, and caches — and guarantees
+  # every match contains '/projects/' for the profile-routing case below.
   local file
-  file=$(command find "$HOME/.claude" "$HOME/.claude-ant" "$HOME/.agents" \
+  file=$(command find "$HOME/.claude/projects" "$HOME/.claude-ant/projects" "$HOME/.agents/projects" \
            -type f -name "${id}.jsonl" -print 2>/dev/null | head -1)
   [[ -n $file ]] || { echo "claude::resume: no session '$id' under ~/.claude*, ~/.agents" >&2; return 1; }
 
PATCH
```

### Verification (of the prepared fix)

- **Syntax**: `zsh -n modules/claude/aliases.zsh` — clean
- **Functional**: scoped `find` resolves a real transcript; `${file%%/projects/*}` → `~/.claude` routes to the default-profile case; `cwd` extraction unchanged
- **Format**: `shfmt` reports pre-existing whole-file reformatting on this zsh-with-bash-shebang file — not introduced here, left untouched

Verdict: comment

PR #33, opened 2026-06-23
fix(durable): green-● and reap broke because ssh truncated the live-port list

DJRHails:
## Root cause

`live_ports` is a space-separated string passed to the remote over ssh:

```sh
ssh "$host" sh -s -- --stream "$host" "$model" "$ttl" "$par" "$lp"
```

ssh flattens its argv into one command string the **remote** shell re-splits, so the quoted-locally `"$lp"` arrives as *N* separate args. The remote reads `live_ports="${6:-}"` — i.e. only the **first** port. After `sort -u` that is always the lowest port (`60001`). Proven: with a 4-port list the remote reports `argc=9`, `$6=60001`, and `$7/$8/$9` hold the rest (ignored).

Two symptoms, one cause:

- **Green ●** — `compute_connf` matched only port `60001`, so the picker always showed exactly **1 connected** (the `60001` session) no matter how many were truly attached. Reproduced live: `$connf` on the host held a single entry.
- **Reap killing sessions** — `mosh_reap` saw a 1-element live list and treated every *other* mosh-server as orphaned, SIGTERM-ing them and dropping the tabs. This is the actual reason `--reap` was nuking sessions.

## Fix

Quote the port list with zsh `${(q)...}` at both ssh call sites (stream + reap) so the whole list survives as one remote arg (`argc` back to 6, `$6` holds the full list). Verified end-to-end: the stream now writes **17 connected** instead of **1**.

## Defense-in-depth (also in this PR)

Even with the list intact, absence-based reaping is fragile, so `mosh_reap` is hardened:

- **Single source of truth** — orphan detection in `orphan_port_pids`, empty-input guard baked in (empty list ⇒ no orphans, fail-safe).
- **Blast-radius cap** — refuse to reap ≥half the servers in one pass (catches partial staleness). This alone would have stopped the truncation bug from mass-killing.
- **Explicit `--force` / `DURABLE_REAP_FORCE`** for deliberate mass cleanup; call site refuses early when this host sees no client.

All guard paths unit-tested; shellcheck + `zsh -n` clean.

## Known residual (not fixed here)

De-nested surfaces whose mosh-server runs a bare `zsh -l` (not `zellij attach cmux-…`) don't map to a session in `compute_connf`'s `pmap`, so they never get a ● even when connected (~1–2 sessions). Tracked separately.

> Note: an unrelated `feat(notify)` commit was pushed onto this branch by a concurrent session; it is not part of the durable fix.

### DIFF
--- a/modules/claude/hooks/notify.sh
+++ b/modules/claude/hooks/notify.sh
@@ -1,8 +1,50 @@
 #!/bin/bash
-# Hook: Notify when Claude needs attention
+set -euo pipefail
+# Hook (Notification): tell me Claude needs attention, with a clickable banner
+# that focuses this session's cmux tab.
+#
+#   - title    = the session's /rename name (resolved from <config>/sessions),
+#                else the project dir name.
+#   - subtitle = the reason Claude surfaced (the hook's .message).
+#   - click    = focus this session's cmux workspace tab (cmux-focus), when the
+#                hook fired inside a cmux surface.
+#
+# Safe + non-blocking: degrades to a plain banner outside cmux or before /rename;
+# never fails the Notification event.
 
-. "$DOTFILES/scripts/core/main.sh"
+INPUT=$(cat)
 
-platform::notify "Claude Code" "Claude needs your attention"
+DOTFILES="${DOTFILES:-$HOME/.files}"
+NOTIFY="$DOTFILES/bin/notify"
+FOCUS="$DOTFILES/bin/cmux-focus"
 
+MESSAGE=$(jq -r '.message // "Claude needs your attention"' <<<"$INPUT")
+SESSION_ID=$(jq -r '.session_id // empty' <<<"$INPUT")
+TRANSCRIPT=$(jq -r '.transcript_path // empty' <<<"$INPUT")
+CWD=$(jq -r '.cwd // empty' <<<"$INPUT")
+
+# Resolve the session's display name (same source as sync-cmux-tab.sh): the
+# sessions file whose .sessionId matches, .name present only after /rename.
+NAME=""
+if [[ -n "$SESSION_ID" && -n "$TRANSCRIPT" ]]; then
+  SESSIONS_DIR="$(dirname "$(dirname "$(dirname "$TRANSCRIPT")")")/sessions"
+  [[ -d "$SESSIONS_DIR" ]] && NAME=$(jq -r --arg sid "$SESSION_ID" \
+    'select(.sessionId==$sid) | .name // empty' "$SESSIONS_DIR"/*.json 2>/dev/null | head -1)
+fi
+PROJECT=""
+[[ -n "$CWD" ]] && PROJECT=$(basename "$CWD")
+TITLE="Claude · ${NAME:-${PROJECT:-Code}}"
+
+# Body carries the project when we have a session name (avoids redundancy).
+BODY="$MESSAGE"
+[[ -n "$NAME" && -n "$PROJECT" ]] && BODY="$MESSAGE — 📁 $PROJECT"
+
+args=(--title "$TITLE" --subtitle "needs your attention" --message "$BODY" --id "claude-$SESSION_ID")
+
+# Clickable focus only when inside a cmux surface and the helper exists.
+if [[ -n "${CMUX_WORKSPACE_ID:-}" && -x "$FOCUS" ]]; then
+  args+=(--button "Focus tab" --run "$FOCUS ${CMUX_WORKSPACE_ID}")
+fi
+
+"$NOTIFY" "${args[@]}" || true
 exit 0
--- a/modules/zellij/durable-remote.sh
+++ b/modules/zellij/durable-remote.sh
@@ -199,24 +199,61 @@ compute_connf() {  # $1 = space-separated host ports with a live local mosh clie
   done | sort -u > "$connf.tmp" && mv -f "$connf.tmp" "$connf"
 }
 
-# Reap mosh-servers with no live local client (their port isn't in $1) that are old enough to be
-# sure. Safe: it only drops the stale transport; the zellij session persists and the picker
-# re-moshes a fresh one. Caveat: "no local client" is judged from THIS machine, so a session
-# you're attached to from another host looks orphaned here.
-mosh_reap() {  # $1 = space-separated host ports with a live local mosh client
-  command -v ss >/dev/null 2>&1 || { printf 'mosh-reap: ss unavailable\n' >&2; return 0; }
-  age_min=120; lp=" ${1:-} "
-  reaped=$(mosh_port_pids | { n=0; while read -r port pid; do
-      case "$lp" in *" $port "*) continue ;; esac   # has a local client → keep
-      a=$(ps -o etimes= -p "$pid" 2>/dev/null | tr -d ' ')
-      [ -n "$a" ] && [ "$a" -gt "$age_min" ] 2>/dev/null && kill -TERM "$pid" 2>/dev/null && n=$((n + 1))
-    done; echo "$n"; })
-  printf 'mosh-reap: %s orphans reaped\n' "$reaped" >&2
+# Ports (and owning pids) of mosh-servers with NO live local client — port absent from $1. The
+# empty-input guard is load-bearing: an empty $1 means we CANNOT SEE any local client (reap run
+# from the wrong host, a picker-teardown race, or a transient ps glitch) — NOT that every server is
+# orphaned. Returning nothing on empty input fails safe: we only ever reap a server we can
+# positively show is clientless, never one we merely can't vouch for. Single source of truth — the
+# reap consumes this; do not re-derive "orphaned" anywhere else.
+orphan_port_pids() {  # $1 = ports with a live local mosh client ; $2 = force (1 = assert all orphaned)
+  if [ -z "${1:-}" ]; then
+    [ "${2:-}" = 1 ] || return 0   # zero knowledge, unforced -> nothing is provably orphaned
+    mosh_port_pids; return 0       # forced -> deliberate mass cleanup, treat every server as orphaned
+  fi
+  lp=" $1 "
+  mosh_port_pids | while read -r port pid; do
+    case "$lp" in *" $port "*) continue ;; esac   # has a local client -> keep
+    printf '%s %s\n' "$port" "$pid"
+  done
 }
 
-# --reap <live-ports>: kill mosh-servers with no live local client (those ports aren't in $2).
-# Cleanup only — the picker computes $connf itself from the same port list. See mosh_reap.
-[ "$mode" = "--reap" ] && { mosh_reap "${2:-}"; exit 0; }
+# Reap mosh-servers with no live local client that are old enough to be sure (>age_min). Safe: it
+# only drops the stale transport; the zellij session persists (its server is PPID 1, daemonized)
+# and the picker re-moshes a fresh one. Two guards make absence-based reaping safe: the empty-input
+# guard in orphan_port_pids, and the blast-radius cap below for *partial* staleness (a stale list
+# that still has a few ports orphans most servers at once). $2=1 overrides both for a deliberate
+# mass cleanup.
+mosh_reap() {  # $1 = live local client ports ; $2 = force (1 = override both safety guards)
+  command -v ss >/dev/null 2>&1 || { printf 'mosh-reap: ss unavailable\n' >&2; return 0; }
+  force="${2:-}"
+  if [ -z "${1:-}" ] && [ "$force" != 1 ]; then
+    printf 'mosh-reap: no live local clients visible — refusing to reap (force=1 to override)\n' >&2
+    return 0
+  fi
+  age_min=120
+  orphans=''
+  # collect orphan pids first so we can size the blast radius before killing anything
+  while read -r port pid; do
+    [ -n "$pid" ] || continue
+    a=$(ps -o etimes= -p "$pid" 2>/dev/null | tr -d ' ')
+    [ -n "$a" ] && [ "$a" -gt "$age_min" ] 2>/dev/null && orphans="$orphans $pid"
+  done <<ORPH
+$(orphan_port_pids "${1:-}" "$force")
+ORPH
+  total=$(mosh_port_pids | grep -c . 2>/dev/null) || total=0
+  norph=0; for _p in $orphans; do norph=$((norph + 1)); done
+  if [ "$norph" -gt 0 ] && [ "$total" -gt 1 ] && [ $((norph * 2)) -ge "$total" ] && [ "$force" != 1 ]; then
+    printf 'mosh-reap: would reap %s of %s servers — refusing mass reap (force=1 to override)\n' "$norph" "$total" >&2
+    return 0
+  fi
+  n=0
+  for _p in $orphans; do kill -TERM "$_p" 2>/dev/null && n=$((n + 1)); done
+  printf 'mosh-reap: %s orphans reaped\n' "$n" >&2
+}
+
+# --reap <live-ports> [force]: kill mosh-servers with no live local client (those ports aren't in
+# $2). Cleanup only — the picker derives connected sessions itself from the same port list.
+[ "$mode" = "--reap" ] && { mosh_reap "${2:-}" "${3:-}"; exit 0; }
 
 sessions=$(list_sessions)
 [ -n "$sessions" ] || exit 0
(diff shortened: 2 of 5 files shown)


modules/zellij/durable-remote.sh
```diff
@@ -199,24 +199,61 @@ compute_connf() {  # $1 = space-separated host ports with a live local mosh clie
...
+  orphans=''
+  # collect orphan pids first so we can size the blast radius before killing anything
+  while read -r port pid; do
+    [ -n "$pid" ] || continue
+    a=$(ps -o etimes= -p "$pid" 2>/dev/null | tr -d ' ')
+    [ -n "$a" ] && [ "$a" -gt "$age_min" ] 2>/dev/null && orphans="$orphans $pid"
+  done <<ORPH
+$(orphan_port_pids "${1:-}" "$force")
+ORPH
+  total=$(mosh_port_pids | grep -c . 2>/dev/null) || total=0
```
sage-saint, line 243: **P2 — correctness (fail-open in the blast-radius cap):** `total` is sampled from a *second, independent* `ss` run, separate from the `orphan_port_pids` snapshot used to build `$orphans`. If that second snapshot transiently under-counts to 0 or 1 (an `ss` glitch, or servers torn down in the millisecond between the two calls — `grep -c` also yields `0` on an `ss` failure via the `|| total=0`), the cap guard `[ "$total" -gt 1 ]` short-circuits to false and **every collected orphan is killed with no cap and no message** — exactly the mass-reap this PR exists to prevent.

Reproduced with a stubbed harness: orphan list from a 4-server fleet + a `total` snapshot returning 1 line → cap skipped, 3 servers killed. Deriving both the orphan list and `total` from a single `mosh_port_pids` snapshot holds the cap (refuses).

Fix (pushed): take one snapshot, feed it to `orphan_port_pids` via stdin and count `total` from the same sample.

modules/claude/hooks/notify.sh
```diff
@@ -1,8 +1,50 @@
 #!/bin/bash
-# Hook: Notify when Claude needs attention
+set -euo pipefail
+# Hook (Notification): tell me Claude needs attention, with a clickable banner
+# that focuses this session's cmux tab.
+#
+#   - title    = the session's /rename name (resolved from <config>/sessions),
+#                else the project dir name.
+#   - subtitle = the reason Claude surfaced (the hook's .message).
```
sage-saint, line 8: **P2 — comment rot:** this says `subtitle = the reason Claude surfaced (the hook's .message)`, but the code does the opposite — `--subtitle "needs your attention"` is a constant (line 42) and the hook's `.message` (`$MESSAGE`) becomes the notification **body** (`--message "$BODY"`). A reader trusting this header would look for `.message` in the wrong field.

Fix (pushed): split the bullet into `subtitle = a constant` and `body = the hook's .message (+ 📁 project)`.

sage-saint (commented):
Automated review (re-review after the `${(q)}` quoting fix). The headline ssh-truncation bug is fixed; two findings remain, posted inline and fixed in a follow-up commit. Each thread is resolved as its fix lands.

modules/zellij/durable-remote.sh
sage-saint: ⚠️ Correction: the fix is prepared and verified locally (commit `41c9434`) but **could not be pushed** — this review worker's token is read-only on contents (`push: false`). The single-snapshot fix is in the PR summary as an applyable patch. Leaving this thread open until the fix lands.

modules/claude/hooks/notify.sh
sage-saint: ⚠️ Correction: fix prepared+verified locally (commit `41c9434`) but not pushable (read-only worker token). Patch in the PR summary. Thread left open until it lands.

sage-saint:
## Review Summary

Re-review of `djrhails/durable-reap-guards` after the `${(q)}` quoting fix (`15bb7a9`). Multi-agent pass (code-review / silent-failure / tests / comments / type-design) on the original PR, then a direct read of the new delta.

### Findings

| F1 | P1 | `ssh "$host" … --reap "$reap_lp" "$reap_force"`: ssh flattens its argv into one string the remote shell re-splits, so a **multi-port** `live_ports` list was truncated to its first port and `force` landed in the wrong positional. With 2+ attached mosh sessions, the others looked orphaned and `--force` was silently ignored. Same flaw in the `--stream` path made the green ● mark only one session. | **Fixed by author in `15bb7a9`** (`${(q)}` on both `--reap` and `--stream`). I verified the quoted list + force both survive the ssh re-split (and the empty-forced case). |
| F2 | P2 | `mosh_reap` sized the blast-radius denominator `total` from a **second, independent `ss` snapshot**. A transient under-count (ss glitch, servers gone between calls, or `grep -c → 0` on ss failure) collapsed `total` to ≤1, disabling the cap guard `[ "$total" -gt 1 ]` and letting an **uncapped mass reap proceed silently** — the exact failure mode this PR exists to prevent. Reproduced with a stubbed harness (4-server fleet + glitched `total=1` → 3 servers killed). | **Fixed** (patch below): one `mosh_port_pids` snapshot feeds both the orphan list (via stdin into `orphan_port_pids`) and `total`. |
| F3 | P2 | `notify.sh` header said `subtitle = the hook's .message`, but the code sets a constant subtitle and puts `.message` in the **body**. | **Fixed** (patch below): comment corrected. |
| F4 | P4 | Cap numerator counts only aged orphans while the denominator counts all servers regardless of age, so the ratio is slightly off from the comment's "fraction of stale servers". Doesn't cause wrong kills (only aged-clientless servers are ever reaped). | Noted, no change — defensible as "fraction of the fleet I'm about to kill". |
| F5 | P3 | The reap guards are safety-critical (`kill -TERM`) but have **no test coverage**; the repo has no shell test harness or CI. F2 (a real fail-open) would have been caught by a unit test. | Recommended, not added (scope). The functions are pure once `ss`/`ps`/`kill` are stubbed — a small `bats`/POSIX harness over `orphan_port_pids`/`mosh_reap` (empty-unforced, empty-forced, cap-refuse, single-reap, force) is cheap. Needs a source-guard (e.g. `[ "${DURABLE_LIB_ONLY:-}" = 1 ] && return 0`) so the dispatch tail doesn't run on source. |
| F6 | P3 | `bin/notify` folds `--open` and `--run` into one `action` passed verbatim to swiftDialog's `--button1action`, with no conflict detection. I could **not verify from this sandbox** (no macOS/swiftDialog) whether `button1action` *runs a command* vs *opens a URL* — the Notification hook depends on `--run "$FOCUS …"` actually executing. | Flagged for confirmation on macOS. If `button1action` opens URLs only, `--run` won't fire the focus handler. |

### Verification

- **shellcheck**: clean — `modules/zellij/durable-remote.sh` (`-s sh`) and `modules/claude/hooks/notify.sh` (`-s bash`).
- **Functional**: extracted the real `mosh_port_pids`/`orphan_port_pids`/`mosh_reap` with `ss`/`ps`/`kill` stubbed and ran all guard branches — empty-unforced→refuse, empty-forced→reap-all, 1-live-of-3→cap refuses, 1-orphan-of-4→reaps the single orphan, force→overrides cap. Behavior preserved; the F2 fix holds the cap when the (formerly second) snapshot would have glitched low.
- **shfmt**: the only diffs are pre-existing house style (compact one-liners / comment spacing) on lines this PR didn't touch; not enforced (no CI / pre-commit / Makefile).
- **No external second opinions** (Codex/Gemini): CLIs not available in this sandbox.

### ⚠️ Fixes prepared but not pushed

This review worker's token is **read-only on contents** (`push: false`; `git push` → 403), so I could post the review but **could not push** the fix commit. The verified fix (local commit `41c9434`) is below — apply from the repo root:

<details><summary><code>git apply</code> the F2 + F3 fixes (one command)</summary>

````sh
git apply <<'PATCH'
diff --git a/modules/claude/hooks/notify.sh b/modules/claude/hooks/notify.sh
index c7d9516827fa..a9a7aca225b9 100755
--- a/modules/claude/hooks/notify.sh
+++ b/modules/claude/hooks/notify.sh
@@ -5,7 +5,9 @@ set -euo pipefail
 #
 #   - title    = the session's /rename name (resolved from <config>/sessions),
 #                else the project dir name.
-#   - subtitle = the reason Claude surfaced (the hook's .message).
+#   - subtitle = a constant ("needs your attention").
+#   - body     = the reason Claude surfaced (the hook's .message), plus 📁 <project>
+#                when a session name is set.
 #   - click    = focus this session's cmux workspace tab (cmux-focus), when the
 #                hook fired inside a cmux surface.
 #
diff --git a/modules/zellij/durable-remote.sh b/modules/zellij/durable-remote.sh
index 67b22b91d87f..7b3977e33896 100755
--- a/modules/zellij/durable-remote.sh
+++ b/modules/zellij/durable-remote.sh
@@ -199,19 +199,22 @@ compute_connf() {  # $1 = space-separated host ports with a live local mosh clie
   done | sort -u > "$connf.tmp" && mv -f "$connf.tmp" "$connf"
 }
 
-# Ports (and owning pids) of mosh-servers with NO live local client — port absent from $1. The
-# empty-input guard is load-bearing: an empty $1 means we CANNOT SEE any local client (reap run
-# from the wrong host, a picker-teardown race, or a transient ps glitch) — NOT that every server is
-# orphaned. Returning nothing on empty input fails safe: we only ever reap a server we can
-# positively show is clientless, never one we merely can't vouch for. Single source of truth — the
-# reap consumes this; do not re-derive "orphaned" anywhere else.
-orphan_port_pids() {  # $1 = ports with a live local mosh client ; $2 = force (1 = assert all orphaned)
+# Ports (and owning pids) of mosh-servers with NO live local client — port absent from $1. Reads a
+# mosh_port_pids snapshot on STDIN (not by re-sampling) so the caller derives both the orphan list
+# and the fleet total from one ss sample — no TOCTOU between two snapshots. The empty-input guard is
+# load-bearing: an empty $1 means we CANNOT SEE any local client (reap run from the wrong host, a
+# picker-teardown race, or a transient ps glitch) — NOT that every server is orphaned. Returning
+# nothing on empty input fails safe: we only ever reap a server we can positively show is clientless,
+# never one we merely can't vouch for. Single source of truth — the reap consumes this; do not
+# re-derive "orphaned" anywhere else.
+orphan_port_pids() {  # $1 = ports with a live local mosh client ; $2 = force ; stdin = mosh_port_pids snapshot
   if [ -z "${1:-}" ]; then
     [ "${2:-}" = 1 ] || return 0   # zero knowledge, unforced -> nothing is provably orphaned
-    mosh_port_pids; return 0       # forced -> deliberate mass cleanup, treat every server as orphaned
+    cat; return 0                  # forced -> deliberate mass cleanup, treat every server as orphaned
   fi
   lp=" $1 "
-  mosh_port_pids | while read -r port pid; do
+  while read -r port pid; do
+    [ -n "$port" ] || continue
     case "$lp" in *" $port "*) continue ;; esac   # has a local client -> keep
     printf '%s %s\n' "$port" "$pid"
   done
@@ -231,6 +234,10 @@ mosh_reap() {  # $1 = live local client ports ; $2 = force (1 = override both sa
     return 0
   fi
   age_min=120
+  # One mosh-server snapshot feeds BOTH the orphan list and the fleet total, so the blast-radius cap
+  # can't be silently disabled by a transient ss glitch collapsing `total` to <=1 between two calls.
+  all_pids=$(mosh_port_pids)
+  total=$(printf '%s\n' "$all_pids" | grep -c . 2>/dev/null) || total=0
   orphans=''
   # collect orphan pids first so we can size the blast radius before killing anything
   while read -r port pid; do
@@ -238,9 +245,8 @@ mosh_reap() {  # $1 = live local client ports ; $2 = force (1 = override both sa
     a=$(ps -o etimes= -p "$pid" 2>/dev/null | tr -d ' ')
     [ -n "$a" ] && [ "$a" -gt "$age_min" ] 2>/dev/null && orphans="$orphans $pid"
   done <<ORPH
-$(orphan_port_pids "${1:-}" "$force")
+$(printf '%s\n' "$all_pids" | orphan_port_pids "${1:-}" "$force")
 ORPH
-  total=$(mosh_port_pids | grep -c . 2>/dev/null) || total=0
   norph=0; for _p in $orphans; do norph=$((norph + 1)); done
   if [ "$norph" -gt 0 ] && [ "$total" -gt 1 ] && [ $((norph * 2)) -ge "$total" ] && [ "$force" != 1 ]; then
     printf 'mosh-reap: would reap %s of %s servers — refusing mass reap (force=1 to override)\n' "$norph" "$total" >&2
PATCH
````

</details>

`request_changes` reflects only that F2/F3 aren't yet landed on the branch (worker can't push); the code itself, after this patch, is sound. The inline threads are left open until the fix lands.


> **Superseded:** verdict downgraded to `comment` — no P1 outstanding (the ssh-truncation P1 was fixed in 15bb7a9; e6e070 is clean). The remaining F2/F3 are non-blocking P2s with a ready patch.

Verdict: comment


_[via gantry](https://gantry.hails.info/run/azure-tranquil-pine)_

sage-saint:
## Delta re-review — `e6e070` (sweep-husks)

Reviewed the new commit `e6e070` *(fix(zellij): make sweep-husks see through the ps-ao wedge and kill servers directly)* against the previously-reviewed `15bb7a9`. One file, 7 lines, in `zellij::sweep-husks()` — a different function from the reap path. **Clean — no new findings.**

Both changes are correct and well-reasoned:

1. **`ps -ao ppid` added to the machinery-exclusion regex** (line 402). zellij 0.44.3 spawns a `ps -ao ppid,args` helper per server; without excluding it, every idle husk would show that helper as a "meaningful" descendant and never be swept. The substring pattern matches regardless of leading path. Correct.
2. **`kill "$spid"` instead of `zellij delete-session --force`** (line 414). Verified `$spid` is the zellij **server** PID (`ps -axww -o pid,command | grep 'zellij --server' | … read -r spid rest`), and the current session (`$agent`) and any session with meaningful descendants are still protected above — so only idle husks are targeted. SIGTERM returns immediately rather than blocking on the wedged proc table that hangs `delete-session`; the cosmetic EXITED entry is pruned by zellij. Strictly better failure mode.

**Verified:** `zsh -n` syntax clean; dry-run path unchanged (no kills); kill target confirmed.

### Still open from the prior review
`e6e070` is a sibling of my (unpushable) fix commit, so **F2 and F3 remain unaddressed on the branch**:
- **F2 (P2)** — `mosh_reap` blast-radius cap can be silently disabled by a transient second-`ss`-snapshot under-count (fail-open mass reap).
- **F3 (P2)** — `notify.sh` header still mislabels `subtitle`/`.message`.

The `git apply` patch in [my earlier summary](https://github.com/DJRHails/dotfiles/pull/33#issuecomment-4780920559) still applies cleanly on `e6e070` (both files untouched by this commit).


> **Superseded:** verdict downgraded to `comment` — no P1 outstanding (the ssh-truncation P1 was fixed in 15bb7a9; e6e070 is clean). The remaining F2/F3 are non-blocking P2s with a ready patch.

Verdict: comment

_[via gantry](https://gantry.hails.info/run/azure-tranquil-pine)_

sage-saint:
## Review status correction — verdict → `comment`

Investigated the `action_required` state on the **Sage Saint Review** check. It's not a broken pipeline — the check concluded correctly from my previous `request_changes` call, and its `details_url` points back at this review run. To re-evaluate it honestly rather than just clear it:

**Re-assessing against my own severity ranking** (only **P1** blocks merge):

- The single P1 I found — `${(q)}`-less ssh arg-flattening that truncated the multi-port `live_ports` list and misplaced `force` — was **fixed by the author in `15bb7a9`** (verified).
- The latest commit `e6e070` (sweep-husks) is **clean**.
- The two findings still open on the branch are both **P2, not merge-blockers**:
  - **F2** — `mosh_reap` blast-radius cap can be bypassed only by a *transient* second-`ss`-snapshot under-count (requires `ss` to glitch to ≤1 between two sub-second calls); the common triggers are already eliminated by the empty-input guard, opt-in `--reap`, and the quoting fix. Real defensive hardening, low probability.
  - **F3** — a `notify.sh` header comment mislabels `subtitle`/`.message`.

Both have a ready one-command `git apply` patch in [the summary](https://github.com/DJRHails/dotfiles/pull/33#issuecomment-4780920559) (still applies cleanly on `e6e070`). I can't land them myself — this review worker's token is read-only on contents (`push: false`).

With no P1 outstanding, `request_changes` was too strict. Downgrading to **`comment`**: the findings stay documented in their open inline threads for the author to apply at leisure, but they don't block merge.

Verdict: comment

_[via gantry](https://gantry.hails.info/run/azure-tranquil-pine)_

PR #38, opened 2026-07-06
fix(hooks): sync-cmux-tab must never block the prompt — fail open on any error

DJRHails:
A non-zero UserPromptSubmit hook **blocks the prompt** — including automated wakes (background-task notifications). In headless gantry workers (djrhails-dev image, no cmux) the empty sessions `*.json` glob reached jq unexpanded, failed the pipeline under `pipefail`, and `set -e` killed the hook with stderr already silenced (`2>/dev/null`) — Claude Code reported *"blocked by hook … No stderr output"* and every background-task wake became an empty session-started/turn-completed pair. A worker parked waiting on a background task could never be woken by its completion (observed on gantry `thrd_gilded-neat-hawk`, 2026-07-06).

Two changes:
- `|| true` on the known-failing `NAME=$(jq … | head -1)` pipeline (empty glob → jq error → pipefail).
- `set -E` + `trap 'exit 0' ERR` so **any** unexpected failure is a silent no-op, making the hook's long-stated contract ("always exits 0 so it never blocks the prompt") structural instead of aspirational.

Verified: reproduced the container failure locally (old hook exits 2 on an empty sessions glob with lib present; fixed hook exits 0 on the same input); shellcheck clean.

### DIFF
--- a/modules/claude/hooks/sync-cmux-tab.sh
+++ b/modules/claude/hooks/sync-cmux-tab.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
-set -euo pipefail
+set -Eeuo pipefail
+# A non-zero UserPromptSubmit hook BLOCKS the prompt — including automated wakes
+# (background-task notifications), which stranded headless gantry workers when
+# this hook died under set -e. Fail open: any unexpected error is a silent no-op.
+trap 'exit 0' ERR
 # Hook (UserPromptSubmit + Stop): keep the cmux panel (the per-surface tab in a
 # pane's tab bar) in step with the Claude session name (set by /rename). cmux
 # already syncs the *workspace* name; the panel/surface tab is not, so we do it.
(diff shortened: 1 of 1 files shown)


DJRHails:
## Review Summary

Post-merge review (PR was already merged when the review ran) — direct single-pass mode: one file, ~10 changed lines.

### Findings

No P1–P3 findings. The fix is correct and well-scoped.

| 1 | P4 | The `|| true` on the `jq … | head -1` pipeline is technically redundant now that `set -E` + `trap 'exit 0' ERR` also covers failures inside command substitutions. | No change: it's deliberate, documented defense-in-depth pinned to the exact known failure — keeping it makes the root cause legible at the call site. |

### Verification (all run locally against this commit)

- **Bug reproduction**: old hook (`985e0463^`) exits **2** on an empty `sessions/*.json` glob with the lib present — reproduces the silent prompt-block. Fixed hook exits **0** on the identical input.
- **ERR trap does structural work beyond the `|| true`**: with an unwritable `$XDG_STATE_HOME` (so `mkdir -p` fails — a path the `|| true` doesn't cover), old hook exits **1**, fixed hook exits **0**. The fail-open contract is now real for arbitrary failures, not just the empty-glob one.
- **Normal paths unchanged**: non-matching session and matching-session-without-cmux both exit 0.
- **Lint/format**: `shellcheck` clean; `shfmt -i 2 -d` clean.

Verdict: approve

_[via gantry](https://gantry.hails.info/run/jaunty-mighty-ash)_

DJRHails:
Acknowledged — thanks for the post-merge review.

Since the only finding (P4, the `|| true` on the `jq … | head -1` pipeline) resolves to **No change** — it's deliberate, documented defense-in-depth pinned to the known empty-glob failure — there are no code changes to make. I re-verified that the merged hook on `main` (985e0463dea1) matches exactly what the review approved: `set -Eeuo pipefail`, `trap 'exit 0' ERR`, and the commented `|| true` at the call site are all in place. Nothing to rebase or push; the PR stays as merged.

_[via gantry](https://gantry.hails.info/run/golden-olive-bison)_

PR #39, opened 2026-07-06
feat(pi): current package name + web-access/subagents/fff extensions

DJRHails:
Switches the pi module to the current `@earendil-works/pi-coding-agent` package (the `@mariozechner` name is deprecated on npm) and adds three extensions to the managed set, in both `install.sh` and the declarative `settings.json`:

- `npm:pi-web-access`
- `npm:pi-subagents`
- `npm:@ff-labs/pi-fff`

Groundwork for gantry's pi worker harness: the worker image bootstraps this module so `pi` + extensions ride the dotfiles, not ad-hoc installs.

PR #40, opened 2026-07-06
fix(pi): migrate existing installs off deprecated @mariozechner package

DJRHails:
Follow-up to #39, resolving review finding F1 (https://github.com/DJRHails/dotfiles/pull/39#discussion_r3531808040).

The package rename in #39 only took effect on machines where `pi` wasn't installed yet: anywhere the deprecated `@mariozechner/pi-coding-agent` was already present, `platform::command_exists "pi"` short-circuited to `log::success`, pinning that machine to the dead package name (frozen at 0.73.1; releases continue only on `@earendil-works/pi-coding-agent`, currently 0.80.3).

This adds a migration branch: when the old global package is detected via `npm ls -g`, uninstall it and install the new name. Fresh installs and already-migrated machines are unaffected (`npm ls -g` exits nonzero → falls through to `log::success`).

**Verification:** `bash -n` and `shellcheck` clean (SC1091 info pre-existing); the `npm ls -g` probe tested both ways — exit 0 for an installed global package, exit 1 for an absent one. Pre-commit hooks (gitleaks, trufflehog, transcrypt check) passed.

_[via gantry](https://gantry.hails.info/run/fair-marble-owl)_

### DIFF
--- a/modules/pi/install.sh
+++ b/modules/pi/install.sh
@@ -4,6 +4,11 @@
 if ! platform::command_exists "pi"; then
   log::execute "npm install -g @earendil-works/pi-coding-agent" \
     "pi-coding-agent"
+elif npm ls -g @mariozechner/pi-coding-agent > /dev/null 2>&1; then
+  # The @mariozechner name is deprecated on npm and receives no new releases.
+  log::execute \
+    "npm uninstall -g @mariozechner/pi-coding-agent && npm install -g @earendil-works/pi-coding-agent" \
+    "pi-coding-agent (migrate to @earendil-works)"
 else
   log::success "pi-coding-agent"
 fi


DJRHails:
## Review Summary

Direct single-pass review (5-line diff, one shell file — follow-up delta to #39, so no agent fan-out).

### Findings

No P1–P3 findings. The migration branch is correct and well-shaped:

- `log::execute` runs its command string through `eval` (`scripts/core/log.sh:59`), so the `npm uninstall … && npm install …` chain executes as intended.
- The probe was re-verified empirically on npm 10.9.8: `npm ls -g <pkg>` exits 0 when the global package is present and 1 when absent, so fresh installs and already-migrated machines fall through to `log::success` as described.
- Failure ordering is self-healing: if the uninstall succeeds but the install fails, `pi` disappears and the next run takes the fresh-install branch, so no machine can get stuck mid-migration.
- If `npm` itself is missing, the `elif` condition fails closed into the success branch — no crash.

P4 (informational, no action): `shfmt` would write `>/dev/null` instead of `> /dev/null`, but both styles coexist across the repo and no formatter hook enforces it.

### Verification

- **Syntax**: `bash -n` clean
- **Lint**: `shellcheck` clean (SC1091 info on the `main.sh` source line is pre-existing and unrelated to this diff)
- **Probe behavior**: `npm ls -g` exit codes tested both ways locally (0 = installed, 1 = absent)
- **Tests/CI**: repo has no CI workflows; pre-commit hooks are secret-scanning only

### Commit

No fix commit — nothing to fix. PR head reviewed: `7ffe462ed006`.

Verdict: approve

_[via gantry](https://gantry.hails.info/run/ancient-cheery-heron)_

PR #41, opened 2026-07-06
fix(pi): sudo the global npm install on Linux (NodeSource system prefix)

DJRHails:
npm -g needs sudo under the NodeSource apt prefix (Linux); without it the pi install fails EACCES inside the gantry worker image bootstrap (unprivileged `dev` user) and the module's extension installs are silently skipped. Same pattern as the node module's `sudo corepack enable pnpm` fallback.

### DIFF
--- a/modules/pi/install.sh
+++ b/modules/pi/install.sh
@@ -1,11 +1,17 @@
 # shellcheck shell=bash
 . "$DOTFILES/scripts/core/main.sh"
 
-if ! platform::command_exists "pi"; then
+# Linux node comes from the NodeSource apt repo (modules/node), whose global
+# prefix is system-owned — npm -g needs sudo there, like the node module's own
+# pnpm fallback. macOS (brew) keeps a user-writable node bin dir.
+if platform::command_exists "pi"; then
+  log::success "pi-coding-agent"
+elif platform::is_osx; then
   log::execute "npm install -g @earendil-works/pi-coding-agent" \
     "pi-coding-agent"
 else
-  log::success "pi-coding-agent"
+  log::execute "sudo npm install -g @earendil-works/pi-coding-agent" \
+    "pi-coding-agent"
 fi
 
 # Install pi extensions


modules/pi/install.sh
```diff
@@ -1,11 +1,17 @@
...
+# prefix is system-owned — npm -g needs sudo there, like the node module's own
+# pnpm fallback. macOS (brew) keeps a user-writable node bin dir.
+if platform::command_exists "pi"; then
+  log::success "pi-coding-agent"
+elif platform::is_osx; then
   log::execute "npm install -g @earendil-works/pi-coding-agent" \
     "pi-coding-agent"
 else
-  log::success "pi-coding-agent"
+  log::execute "sudo npm install -g @earendil-works/pi-coding-agent" \
```
DJRHails, line 13: **P3 — consistency / robustness:** use the repo's `platform::sudo` helper instead of bare `sudo` here.

`platform::sudo` (scripts/core/platform.sh:199) runs `sudo "$@"` only when `EUID -ne 0`, else runs the command directly. Bare `sudo` is the repo's minority idiom — ~15 modules (node's apt path, docker, tailscale, keybase, cloudflared, git…) use `platform::sudo`; the only two bare-`sudo` install sites are this line and the `node` pnpm fallback (`modules/node/install.sh:30`) that this PR deliberately mirrored. So the fix followed the one inconsistent sibling rather than the dominant convention.

Concrete failure: if this module ever bootstraps as root in an image without `sudo` on PATH (common in minimal root containers), `sudo npm ...` fails `command not found` where `platform::sudo` would no-op to a direct `npm -g` that succeeds (root already owns the NodeSource `/usr` prefix). Not a problem in the current gantry image (unprivileged `dev`, sudo present), hence P3.

Suggested: `log::execute "platform::sudo npm install -g @earendil-works/pi-coding-agent" "pi-coding-agent"`. Worth converting the node pnpm fallback in the same sweep to close the bug class.

DJRHails (commented):
Automated review (post-merge) — one P3 finding posted inline. Fix opened as a follow-up PR since this one is already merged.

modules/pi/install.sh
DJRHails: Fixed in 23eeea5efb3a via follow-up #42 (this PR was already merged, so the fix lands on a new branch). Converted both this line and the mirrored `node` pnpm fallback to `platform::sudo`.

DJRHails:
## Review Summary

Post-merge review of #41 (direct single-pass — 1 file, +8/-2). The change is **correct**: it fixes a real silent-skip where `npm -g` hit `EACCES` against NodeSource's system-owned `/usr` prefix as the unprivileged `dev` user, so `pi` never installed and the extension block was skipped. The comment is accurate and `log::execute` surfaces failures (streams stderr on non-zero exit), so the fix genuinely closes the silent path.

### Findings

| 1 | P3 | Bare `sudo` instead of the repo's `platform::sudo` helper (`scripts/core/platform.sh`). `platform::sudo` no-ops when already root; bare `sudo` fails `command not found` if the module ever bootstraps as root in a sudo-less image. This PR mirrored the one inconsistent sibling (`node` pnpm fallback) rather than the dominant idiom (~15 modules use `platform::sudo`). | Fixed in follow-up #42 (`23eeea5`) — converted both `modules/pi/install.sh` and the `modules/node/install.sh` pnpm fallback. |

No P1/P2 issues. No test/type/build gates in this repo (dotfiles; only secret-scan + transcrypt pre-commit hooks).

### Verification

- **Syntax**: `bash -n` clean on both files
- **Lint**: `shellcheck` clean (only pre-existing SC1091 source-follow info)
- **Scope check**: confirmed `platform::sudo` is defined (both files `. main.sh` → `platform.sh`) and survives the `log::execute` eval-in-background-subshell path
- **Hooks**: gitleaks, trufflehog, transcrypt-check all passed

### Fix

Follow-up PR #42 · commit `23eeea5efb3a` — `fix: resolve code review findings for PR #41`

Verdict: approve

_[via gantry](https://gantry.hails.info/run/sharp-mossy-thistle)_

PR #45, opened 2026-07-10
docs(slack): clarify per-workspace posting guardrails

DJRHails:
Documents which workspace the skill's write commands may treat as low-friction versus where the confirm-first guards apply in full. The touched file is encrypted at rest; details are in the (encrypted) file itself.

_[via gantry](https://gantry.hails.info/run/humble-jolly-ledge)_


DJRHails:
## Review Summary

**Review mode:** direct single-pass (one docs file). The only changed file, `modules/agents/skills/slack/SKILL.md`, is transcrypt-encrypted (`filter=crypt` via `.gitattributes:10`) and this reviewer's scoped checkout intentionally has no decryption key — so the prose content was **not** reviewed. What was reviewed is the encryption boundary and blob integrity, which is the substance that can go wrong on a PR like this.

### Findings

| 1 | P4 | Skill prose is unreviewable without the transcrypt key; this review attests boundary integrity only | Informational — by design for encrypted-at-rest files |

No P1–P3 findings; no fix commit needed.

### Verification

- **Encryption boundary:** `git check-attr` confirms `filter/diff/merge=crypt` on the path; both the base (`origin/main`) and head blobs are valid base64 decoding to OpenSSL `Salted__` payloads (12,576 → 13,040 bytes ciphertext). Working tree and committed blob are ciphertext — no plaintext committed.
- **Metadata hygiene:** commit subject, PR title, and PR body stay generic and explicitly defer details to the encrypted file — consistent with the repo's encrypted-at-rest convention.
- **Scope:** exactly one file changed; a repo-wide scan for stray plaintext copies of the skill found only pre-existing, unrelated mentions (a policy string in `modules/claude/settings.ant.json`, no secrets).
- **Tests/Lint/CI:** none applicable — no workflows in `.github/workflows/`, docs-only diff.

### Commit

No fix commit — PR head remains `13eaa7e9099b`.

Verdict: comment

_[via gantry](https://gantry.hails.info/run/silver-woven-owl)_

PR #46, opened 2026-07-10
feat(zellij): pane frames + keybind hints, per-host accent themes

DJRHails:
Soften the fully-quiet zellij look and add per-host accent colours.

- `pane_frames true`, default layout back (tab bar + status bar with keybind hints)
- `theme "host"` + `setup.sh`: links `~/.config/zellij/themes/host.kdl` → `themes/<hostname>.kdl` (`default.kdl` fallback = stock green, so an unthemed host is visibly unthemed)
- Accents: **bonbon = red** (prod, the careful colour) · **taffy = pink** (salt-water taffy) · **tiramisu = coffee** · **trifle = blue** (home base)
- Missing `host.kdl` is non-fatal — zellij 0.44.3 falls back to its default theme (tested live)
- Theme files override only accent components (`frame_selected`, `frame_highlight`, `ribbon_selected`, `table_title`, `table_cell_selected`, `list_selected`); everything else inherits zellij's `DEFAULT_STYLES` per-component fallback

PR #47, opened 2026-07-10
feat: Option-as-Alt (ghostty) + stacked-by-default panes (zellij)

DJRHails:
Follow-up to #46, two small quality-of-life fixes:

- **zellij**: `Alt n` now opens panes **stacked** (title-bar strips, focused pane expanded) instead of tiling. Directional splits remain on `Ctrl p d`/`r`. Keybind block merges with defaults — only `Alt n` overridden. Verified on 0.44.3 with synthetic `Alt+n` in a scratch session.
- **ghostty**: `macos-option-as-alt = left` — macOS was composing special characters on Option, so zellij never saw any Alt binds. Left Option = Alt; right Option keeps character composition (#, dead keys).

### DIFF
--- a/modules/ghostty/config/ghostty/config
+++ b/modules/ghostty/config/ghostty/config
@@ -5,3 +5,9 @@
 # Tradeoff: Shift+drag no longer extends selection *inside* the multiplexer —
 # native terminal selection takes over instead.
 mouse-shift-capture = never
+
+# Left Option sends Alt (ESC-prefixed), so zellij's Alt binds (Alt n new pane,
+# Alt hjkl focus, Alt +/- resize) work. `left` not `true`: the RIGHT Option key
+# keeps macOS character composition, so Option-composed symbols (#, €, dead
+# keys) still type via right Option.
+macos-option-as-alt = left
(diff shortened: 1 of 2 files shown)


DJRHails:
## Review Summary

Post-merge review (PR merged before the review ran). Direct single-pass review — 32 diff lines across two config files, no source logic.

### Findings

No P1–P3 findings. Both changes are correct and well-documented:

- **zellij** `keybinds { shared_except "locked" { bind "Alt n" { NewPane "stacked"; } } }` — valid KDL, merges with the default binds (no `clear-defaults`), and overrides `Alt n` in the same scope the default uses. Comment matches actual behavior (directional splits remain on `Ctrl p` → `d`/`r`).
- **ghostty** `macos-option-as-alt = left` — valid value; `left` correctly preserves right-Option character composition. macOS-prefixed key is a no-op on Linux hosts, so the shared config stays portable.

### Verification

- Downloaded the exact zellij **0.44.3** binary pinned by `modules/zellij/install.ubuntu.sh` (checksum verified) and ran `zellij setup --check` against this repo's `config.kdl` (with the `host` theme symlink mimicked): **config parses clean, `NewPane "stacked"` accepted**.
- Confirmed the check has teeth: replacing `"stacked"` with a bogus direction fails parsing (`Invalid direction`), so the pass is meaningful.
- No repo CI to run; no shell/source files changed.

### Commit

No fix commits — nothing to fix.

Verdict: approve

_[via gantry](https://gantry.hails.info/run/calm-coral-reed)_

PR #51, opened 2026-07-11
docs(skills): refresh tailscale-topology host inventory

DJRHails:
Refreshes the (transcrypt-encrypted) tailnet inventory skill — one node's entry was stale after it re-registered, plus a previously undocumented host row. Details are inside the encrypted file.


DJRHails:
## Review Summary

Post-merge review (the PR merged as `704bd8c9` before this review ran). The changed file is transcrypt-encrypted and this reviewer runs without the decryption key, so the review covered encryption integrity and plaintext metadata rather than the inventory content itself.

### Findings

| 1 | P3 | The head branch name embeds a tailnet hostname that appears nowhere else in the repo's plaintext — the exact class of detail the encrypted file exists to hide, and it identifies which node the deliberately-generic PR body refers to. The remote branch is already deleted, but GitHub retains a merged PR's head branch name permanently, so it cannot be scrubbed. | Prevention landed: [PR #52](https://github.com/DJRHails/dotfiles/pull/52) extends the encrypted-at-rest Git Hygiene rule in `modules/agents/AGENTS.md` to cover branch names (they must be generic *before* the first push, since there is no after-the-fact fix). |

No inline comments: the diff is ciphertext, and the single finding is about PR metadata, not a diff line.

### What was verified

- Committed blob at `modules/agents/skills/tailscale-topology/SKILL.md` is proper transcrypt ciphertext (`Salted__` header) — no plaintext escaped into the commit.
- Commit message, PR title, and PR body are appropriately generic — no inventory details leaked there.
- Hostname sweep across all tracked plaintext files and commit messages confirmed the branch-name disclosure is the only new exposure.
- `scripts/check-crypt-patterns.sh`: **pass**. `gitleaks` (staged) and `trufflehog` (filesystem): **no leaks** on the fix change.

### Commit

Fix pushed as `e3ab932` on [PR #52](https://github.com/DJRHails/dotfiles/pull/52) — `docs(agents): extend encrypted-at-rest rule to branch names`.

Verdict: comment

_[via gantry](https://gantry.hails.info/run/cool-quirky-bay)_

PR #52, opened 2026-07-11
docs(agents): extend encrypted-at-rest rule to branch names

DJRHails:
The Git Hygiene rule for encrypted-at-rest files enumerates the plaintext metadata surfaces to keep generic (commit message, PR title/description, issue text, review comments) but omitted **branch names**. #51's head branch name embedded a detail its encrypted file exists to hide — and branch names are the one surface with no after-the-fact fix, since GitHub retains a PR's head branch name even after the branch is deleted.

This adds branch names to the enumeration and a note that they must be chosen generically *before* the first push.

Found during the post-merge review of #51 ([run](https://gantry.hails.info/run/cool-quirky-bay)).


DJRHails:
## Review Summary

Direct single-pass review (docs-only, +1/−1) — no sub-agent fan-out needed.

### Findings

No P1–P3 findings. The change is accurate and well-scoped:

- GitHub does retain a PR's head branch name permanently (API `headRefName` and the PR page survive branch deletion), so "no after-the-fact fix" is correct and properly scoped to PRs.
- The `topology-<hostname>` example uses a placeholder and "topology" is already public in main's plaintext history, so the example doesn't itself leak.
- Only one copy of the rule exists in the repo (`modules/agents/AGENTS.md`), and the file is not transcrypt-filtered, so editing it in plaintext is fine.
- The only other branch-naming guidance in the repo (`fix-issue.md`'s `issue-N` pattern) is generic by construction — no contradiction introduced.
- "Before pushing" is stricter than "before opening a PR", but correctly so: pushed refs leak via the events API even without a PR.

### Verification

- **Tests / Lint / Format**: n/a — markdown-only change; repo has no CI workflows or Makefile.
- Checked for duplicate rule copies (`rg "Encrypted-at-rest files"`) and encryption attrs (`git check-attr filter`).

### Commit

No fix commit — nothing to fix.

Verdict: approve

_[via gantry](https://gantry.hails.info/run/lush-upbeat-meadow)_

PR #56, opened 2026-07-11
fix: resolve review findings for PR #54 — shell startup caching + zellij GC

DJRHails:
Follow-up to #54, which merged before its review concluded. Fixes all nine inline findings from the [review](https://github.com/DJRHails/dotfiles/pull/54#pullrequestreview-4678821298).

## Headline (P1)

**`zellij-gc` never ran in production.** `(#q...)` glob qualifiers require `extended_glob`, which is off under `zsh -f` (how auto-attach spawns it), so `[[ -n $throttle_stamp(#qN.mh-6) ]]` was a literal non-empty string — always true — and every flagless run exited at the throttle before reaping anything. `--dry-run`/`--force` bypass that check, which is why manual testing looked healthy. Reproduced on zsh 5.9 with a fresh stamp, a 10h-old stamp, and no stamp at all. Fix: `setopt extended_glob`.

## Also fixed

- **compinit gate** (P2): same `extended_glob` bug made `compinit -C` dead code — full compinit ran every shell. Now an anonymous function with `localoptions extendedglob`, inverted polarity (fresh dump → `-C`), and a `touch` after full compinit so the 24h clock actually advances.
- **zprofile brew caching** (P2): `brew shellenv` prints ~nothing when `HOMEBREW_PREFIX` is already set, so a nested login shell could cache an empty body and starve every fresh login shell of brew env. Whole block now guarded on `HOMEBREW_PREFIX` being unset.
- **`_cached_eval` `emulate -L`** (P2): localized `setopt`/`typeset` from sourced init output (starship emits `setopt promptsubst` — currently masked only by config.zsh's independent `PROMPT_SUBST`). Dropped; all vars were already `local`.
- **gc EXITED sweep** (P2): missing stamp deleted records immediately, and a reap + delete could destroy resurrectable state in one pass. Reaping now resets the stamp (7-day window starts at exit); missing stamps get seeded, never immediate-deleted. Detach in auto-attach also refreshes the stamp.
- **gc busy check** (P3): direct non-shell server children (`zellij run -- cmd`, exec'd programs) counted as idle; now they count busy.
- **silent missing sheldon** (P3): now warns instead of silently skipping all plugins.
- **gc robustness** (P3): non-numeric `CMUX_ZELLIJ_GC_STALE_HOURS` falls back to 48 (previously errored into reaping *fresh* sessions); `--dry-run` no longer writes gc.log; `run complete reaped=N deleted=N` breadcrumb.
- **`_cached_eval` cache-fill** (P3): tool stderr passes through instead of being swallowed into a persistently-cached degraded state; header now records argv so distinct invocations can't collide on one cache file.

## Tests

#54's description claimed a behavior suite that was never committed. Now in-tree, self-contained (`zsh -f tests/<name>.test.zsh`):

- `tests/cached-eval.test.zsh` — cold/warm/binary-upgrade/dep-invalidation/corrupt-cache/read-only dir/missing binary/setopt-escape/6-way concurrent cold start. **12/12 pass.**
- `tests/zellij-gc.test.zsh` — throttle first-run/fresh/stale, dry-run purity, unknown flag, bad stale-hours env. **7/7 pass — and 2 fail against the pre-fix script** (falsification of the P1).

Verification beyond the suites: `zsh -n` on all changed files; full interactive login boot simulated with the changed zshenv/zprofile/zshrc pieces (compinit branch selection, `extendedglob` not leaking, `compdef` defined); busy-check comm classification exercised against live pids; gitleaks + trufflehog clean.

### DIFF
--- a/modules/zellij/zellij-gc
+++ b/modules/zellij/zellij-gc
@@ -11,17 +11,23 @@
 #
 # Only touches sessions named cmux-* . A session is reaped when it is:
 #   - not attached (no `zellij attach <name>` client process), and
-#   - idle (no pane shell has a child process other than ps), and
+#   - idle (every server child is a plain shell, and no pane shell has a
+#     child process other than ps), and
 #   - stale (its auto-attach stamp $logdir/last-<name> is older than
 #     $CMUX_ZELLIJ_GC_STALE_HOURS, default 48; missing stamp counts as stale).
 # Reaped live sessions get SIGTERM — serialized state stays on disk, so they
-# remain resurrectable. EXITED session records older than 7 days are deleted.
+# remain resurrectable — and their stamp is reset so the EXITED sweep below
+# won't delete them for another $exited_max_days. EXITED records are deleted
+# once their stamp is older than 7 days; a record with no usable stamp gets
+# one seeded now, so deletion happens no earlier than 7 days after first
+# sighting.
 #
 # Usage: zellij-gc [--dry-run] [--force]
-#   --dry-run  print what would be reaped, change nothing
+#   --dry-run  print what would be reaped; touches no sessions, stamps, or logs
 #   --force    ignore the 6h throttle stamp
 
 set -u
+setopt extended_glob  # (#q...) glob qualifiers below are inert without this
 zmodload zsh/datetime
 
 dry_run=0 force=0
@@ -37,13 +43,17 @@ logdir="${XDG_CACHE_HOME:-$HOME/.cache}/cmux-zellij"
 gc_log="$logdir/gc.log"
 throttle_stamp="$logdir/gc-last-run"
 stale_hours="${CMUX_ZELLIJ_GC_STALE_HOURS:-48}"
+[[ $stale_hours == <-> ]] || stale_hours=48
 exited_max_days=7
 
 mkdir -p "$logdir" 2>/dev/null || exit 0
 
 log() {
-  print -r -- "$(strftime '%Y-%m-%d %H:%M:%S' $EPOCHSECONDS) $*" >> "$gc_log" 2>/dev/null
-  (( dry_run )) && print -r -- "$*"
+  if (( dry_run )); then
+    print -r -- "$*"
+  else
+    print -r -- "$(strftime '%Y-%m-%d %H:%M:%S' $EPOCHSECONDS) $*" >> "$gc_log" 2>/dev/null
+  fi
 }
 
 # Throttle: bail if we ran within the last 6 hours (unless --force / --dry-run).
@@ -72,9 +82,17 @@ ps -eo pid,args 2>/dev/null | grep 'zellij --server' | grep -v grep | while read
   [[ $sess == cmux-* ]] || continue
   [[ -n ${protected[$sess]:-} ]] && continue
 
-  # busy check: any pane shell with a non-ps child means real work is running
+  # busy check: a direct server child that isn't a plain shell is a command
+  # pane (`zellij run -- cmd`, an exec'd program) — real work even with no
+  # children. For shell children, any non-ps grandchild means a running job.
   busy=0
   for shell_pid in $(pgrep -P "$pid" 2>/dev/null); do
+    child_comm="${$(ps -p "$shell_pid" -o comm= 2>/dev/null):t}"
+    case ${child_comm#-} in
+      zsh|bash|fish|sh|dash) ;;
+      '') continue ;;  # vanished between pgrep and ps; nothing to inspect
+      *) busy=1; break ;;
+    esac
     for job in $(pgrep -P "$shell_pid" 2>/dev/null); do
       [[ "$(ps -p "$job" -o comm= 2>/dev/null)" == ps ]] && continue
       busy=1; break 2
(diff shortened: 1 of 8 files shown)


tests/zellij-gc.test.zsh
```diff
@@ -0,0 +1,61 @@
...
+zsh -f "$gc" --dry-run
+check dry-run-no-stamp "no" "$([[ -e $stamp ]] && print yes || print no)"
+check dry-run-no-log "no" "$([[ -e $logdir/gc.log ]] && print yes || print no)"
+
+# unknown flag: usage error
+zsh -f "$gc" --bogus 2>/dev/null
+check unknown-flag-exit-2 "2" "$?"
+
+# non-numeric stale-hours env must not error (falls back to the default)
+err="$(CMUX_ZELLIJ_GC_STALE_HOURS=abc zsh -f "$gc" --force 2>&1)"
```
DJRHails, line 58: **P2 — test gap (vacuous check):** this check cannot fail. The arithmetic that a non-numeric `CMUX_ZELLIJ_GC_STALE_HOURS` would break (`stale_before=$(( now - stale_hours * 3600 ))`, zellij-gc line 76) sits *after* the `command -v zellij || exit 0` gate at line 65 — and this suite deliberately hides zellij from PATH. Verified: reverting the `[[ $stale_hours == <-> ]]` guard and running this exact scenario still produces empty stderr; with a stub `zellij` on PATH the unguarded script errors with `abc: parameter not set`. Fix: run this check with stub `zellij`/`ps` binaries so execution reaches the arithmetic (stubbing `ps` also keeps the reap loop from ever seeing real servers on a dev machine running the suite).

modules/zellij/zellij-gc
```diff
@@ -91,27 +109,45 @@ ps -eo pid,args 2>/dev/null | grep 'zellij --server' | grep -v grep | while read
...
+      print $EPOCHSECONDS > "$stamp" 2>/dev/null
+      log "seeded stamp for exited sess=$sess"
+    fi
+    continue
   fi
+  (( last > exited_before )) && continue
   if (( dry_run )); then
     log "would-delete exited sess=$sess"
   else
     zellij delete-session "$sess" >/dev/null 2>&1 && log "deleted exited sess=$sess"
```
DJRHails, line 146: **P3 — resource leak:** deleting an EXITED record strands its sidecar files. `zellij delete-session` removes zellij's record, but `$logdir/last-$sess` (and auto-attach's `live-$sess`) stay behind forever — and the new seeding above guarantees every record has a stamp by the time it's deleted, so every deletion now leaks exactly one file into the GC's own cache dir. Monotonic growth in the very directory this script exists to keep bounded. Fix: `rm -f -- "$stamp" "$logdir/live-$sess"` on successful delete.

DJRHails (commented):
Post-merge review — findings posted inline. Each thread is resolved as its fix lands (fixes go to a follow-up PR since this one is merged).

tests/zellij-gc.test.zsh
DJRHails: Fixed in b5935a74669e (PR #57): the check now runs against stub `zellij`/`ps` binaries as `--dry-run` with stderr-only capture, so it reaches the arithmetic — verified it fails with `abc: parameter not set` when the guard is reverted. The stub `ps` also keeps the reap loop from ever seeing real servers on a machine running the suite.

modules/zellij/zellij-gc
DJRHails: Fixed in b5935a74669e (PR #57): successful deletes now `rm -f -- "$stamp" "$logdir/live-$sess"`, with new tests proving both sidecars are gone after the sweep (they fail against the pre-fix script).

DJRHails:
## Review Summary

Post-merge review of #56 (direct single-pass — follow-up delta to #54's full battery; ~110 hand-written source lines). Every changed hunk read against full file context; both suites executed; both findings verified by falsification before posting. Fixes landed via #57 since this PR was already merged.

### Findings

| F1 | P2 | `bad-stale-hours-silent` was vacuous: the failing arithmetic sits after the `command -v zellij \|\| exit 0` gate and the suite hides zellij — reverting the guard still passed | Fixed in [#57](https://github.com/DJRHails/dotfiles/pull/57): check runs against stub `zellij`/`ps` (dry-run, stderr-only); now fails `abc: parameter not set` pre-fix |
| F2 | P3 | EXITED delete strands sidecar files — with #56's seeding, every delete leaks one `last-<sess>` (+ `live-<sess>`) into the GC's own logdir, forever | Fixed in [#57](https://github.com/DJRHails/dotfiles/pull/57): sidecars removed on successful delete; new sweep tests fail against pre-fix script |

P4 (informational, no action): the detach-time stamp refresh can trip the 3s attach circuit-breaker if a new shell spawns in the same surface within 3s of detach (one-shot refusal with clear error, self-heals); `reaped`/`deleted` counters include failed kills; no CI runs these suites (`.github/` absent).

Everything else checked out: the `extended_glob` P1 fix and its falsifying throttle tests, the compinit polarity inversion + `touch`, the `HOMEBREW_PREFIX` zprofile guard, `_cached_eval` restructure (precedence of the `!…||…` chain verified, stderr passthrough, argv-keyed header), and the busy-check comm classification.

### Verification

- **Tests**: `tests/zellij-gc.test.zsh` 10/10 (8 consecutive clean runs), `tests/cached-eval.test.zsh` 12/12; falsification runs: 2 checks fail vs merged main's gc, 3 vs a guard-reverted variant
- **Lint**: `zsh -n` clean on all 6 changed source files
- **Secrets**: gitleaks (staged) + trufflehog (filesystem) clean; those two pre-commit hooks skipped by id (golang bootstrap blocked by GitHub-only egress), `check-crypt-patterns` passed

### Commit

`b5935a74669e` — fix: resolve post-merge review findings for PR #56 — zellij-gc (squash-merged to main as `ffa4962651a1` via #57)

Verdict: comment

_[via gantry](https://gantry.hails.info/run/glad-fair-badger)_

PR #57, opened 2026-07-11
fix: resolve post-merge review findings for PR #56 — zellij-gc sidecar leak + vacuous test

DJRHails:
Follow-up to #56 (merged before this post-merge review concluded). Fixes both inline findings from the [review](https://github.com/DJRHails/dotfiles/pull/56#pullrequestreview-4678859812), each verified by falsification — the new/changed tests fail against the pre-fix script.

## Sidecar-file leak on EXITED delete (P3, finding F2)

`zellij delete-session` removed the record but stranded `$logdir/last-<sess>` (and auto-attach's `live-<sess>`) forever — and #56's stamp seeding guaranteed every record had a stamp by deletion time, so **every delete leaked exactly one file** into the directory the GC exists to keep bounded. Both sidecars are now removed on successful delete.

## Vacuous `bad-stale-hours-silent` test (P2, finding F1)

The arithmetic a non-numeric `CMUX_ZELLIJ_GC_STALE_HOURS` breaks (`stale_before=$(( now - stale_hours * 3600 ))`) sits *after* the `command -v zellij || exit 0` gate, and the suite deliberately hides zellij — so the check passed even with the guard reverted. It now runs against a stub `zellij` (as `--dry-run`, capturing stderr only) so execution reaches the arithmetic; a stub `ps` guarantees the reap loop can never see — or signal — real servers on a machine running the suite.

## New coverage

First tests for the EXITED sweep: a no-stamp record gets seeded (not deleted); a record whose stamp aged past 7 days is deleted **with** its sidecar files.

## Verification

- `zsh -n` on both changed files
- `tests/zellij-gc.test.zsh` **10/10** (8 consecutive clean runs); `tests/cached-eval.test.zsh` **12/12**
- Falsification: against merged main's `zellij-gc` the two sidecar checks fail; against a guard-reverted variant, `bad-stale-hours-silent` additionally fails with `abc: parameter not set` — previously it passed vacuously
- gitleaks (staged) + trufflehog (filesystem) clean

### DIFF
--- a/modules/zellij/zellij-gc
+++ b/modules/zellij/zellij-gc
@@ -142,8 +142,12 @@ for sess in $(zellij list-sessions 2>/dev/null | perl -pe 's/\e\[[0-9;]*m//g' \
   (( last > exited_before )) && continue
   if (( dry_run )); then
     log "would-delete exited sess=$sess"
-  else
-    zellij delete-session "$sess" >/dev/null 2>&1 && log "deleted exited sess=$sess"
+  elif zellij delete-session "$sess" >/dev/null 2>&1; then
+    # The record is gone; drop its sidecar files too (the stamp, plus
+    # auto-attach's live-ids), else every deletion strands one file in the
+    # very directory this script exists to keep bounded.
+    rm -f -- "$stamp" "$logdir/live-$sess" 2>/dev/null
+    log "deleted exited sess=$sess"
   fi
   (( deleted++ ))
 done
--- a/tests/zellij-gc.test.zsh
+++ b/tests/zellij-gc.test.zsh
@@ -1,9 +1,12 @@
 #!/usr/bin/env zsh
-# Throttle and flag behavior for zellij-gc. Needs no zellij binary: the script
-# exits right after the throttle logic when zellij is absent, so whether the
-# throttle stamp got (re)written proves which side of the throttle a run
-# landed on. Regression test for the extended_glob bug that made the throttle
-# always fire — with it, the first run below never writes the stamp.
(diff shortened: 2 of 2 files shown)


DJRHails:
## Review Summary

Post-merge review (PR merged as `ffa4962` before review started). Direct single-pass review — 82-line delta across 2 files, itself a follow-up to the reviewed PR #56.

### Findings

No P1–P3 findings. Both changes verified correct:

- **Sidecar cleanup** (`modules/zellij/zellij-gc:149`): `rm -f -- "$stamp" "$logdir/live-$sess"` on successful `delete-session` — confirmed `live-$sess` is the exact sidecar auto-attach writes (`modules/zellij/auto-attach.zsh:159`; mosh path writes the same name remotely). Removal is correctly gated on delete success, so a failed delete keeps the stamp and the record stays scheduled.
- **Test de-vacuification** (`tests/zellij-gc.test.zsh`): the stub `zellij`/`ps` approach cleanly gets past the `command -v zellij` exit without touching real sessions, and the `bad-stale-hours` check now actually reaches the arithmetic it guards.

### Verification

- **Tests**: `zsh -f tests/zellij-gc.test.zsh` — all 10 checks pass.
- **Mutation checks** (tests catch failures): removing the sidecar `rm` fails `exited-stale-stamp-removed` + `exited-stale-live-removed`; removing the `stale_hours` numeric guard fails `bad-stale-hours-silent` with the exact `bad math expression`-class error the check exists to catch. Both restored; suite green.

No fix commits needed.

Verdict: approve

_[via gantry](https://gantry.hails.info/run/shy-sandy-pebble)_

PR #58, opened 2026-07-15
feat(zsh): auto-load direnv per directory (cached)

DJRHails:
Adds a guarded, `_cached_eval`'d `direnv hook zsh` to the zshrc template so every machine gets per-directory `.envrc` auto-loading on next render.

- Guarded on `$+commands[direnv]` → no-ops on hosts without direnv installed.
- Cached via `_cached_eval` like the starship init line above it.
- Previously direnv was hand-added to the *generated* `zshrc` on bonbon, which broke the `.zshrc.sha256` guard and froze template updates. This moves it to the tracked source.

### DIFF
--- a/modules/zsh/zshrc.tmpl
+++ b/modules/zsh/zshrc.tmpl
@@ -106,3 +106,8 @@ export PATH=${PATH:+$PATH:}${DOTFILES}/bin
 
 # Starship prompt (cached: skips the ~40ms init subprocess; see zshenv)
 (( $+commands[starship] )) && _cached_eval starship init zsh
+
+# direnv: auto-load .envrc (which dotenvs .env) per directory, cached like
+# starship (see zshenv's _cached_eval). Guarded on $+commands so it no-ops
+# on hosts where direnv isn't installed.
+(( $+commands[direnv] )) && _cached_eval direnv hook zsh


modules/zsh/zshrc.tmpl
```diff
@@ -105,4 +105,9 @@ bindkey "^[[B" down-line-or-beginning-search # Down
...
 
 # Starship prompt (cached: skips the ~40ms init subprocess; see zshenv)
-(( $+commands[starship] )) && _cached_eval starship init zsh
\ No newline at end of file
+(( $+commands[starship] )) && _cached_eval starship init zsh
+
+# direnv: auto-load .envrc (which dotenvs .env) per directory, cached like
+# starship (see zshenv's _cached_eval). Guarded on $+commands so it no-ops
+# on hosts where direnv isn't installed.
+(( $+commands[direnv] )) && _cached_eval direnv hook zsh
```
DJRHails, line 113: **P2 — first prompt shows a spurious error on hosts without direnv.** This guard is now the *last* command in the generated zshrc, and when direnv is absent `(( $+commands[direnv] )) && …` short-circuits with status 1. zsh hands that status to the first precmd (verified empirically: a trailing failed `&&` guard makes the first precmd see `$?=1`), and starship — initialized two lines up — captures it as `STARSHIP_CMD_STATUS`, so its prompt character renders the red `error_symbol` on every new shell. Before this PR the final line was the starship guard, which returns 0 whenever starship exists, so this is a regression on starship-but-no-direnv hosts. There is no zlogin in this repo, so nothing runs after zshrc to reset the status.

Fix: use the `if` form — a failed `if` with no `else` returns 0:

```zsh
if (( $+commands[direnv] )); then
  _cached_eval direnv hook zsh
fi
```

DJRHails (commented):
Automated post-merge review — finding posted inline; fix follows as a separate PR since #58 is already merged.

modules/zsh/zshrc.tmpl
DJRHails: Fixed in ac10508b9104 — opened as [PR #59](https://github.com/DJRHails/dotfiles/pull/59) since #58 was already merged. Verified: first precmd now sees status 0 with direnv absent, and both test suites pass.

DJRHails:
## Review Summary

Post-merge review (direct single-pass — 6-line diff to one template file). #58 was already merged when the review ran, so the fix landed as [PR #59](https://github.com/DJRHails/dotfiles/pull/59) off `main` instead of a push to the PR branch.

### Findings

| F1 | P2 | The new direnv guard is the last command in the generated zshrc; when direnv is absent, `(( $+commands[direnv] )) && …` leaves status 1, which the first precmd observes and starship renders as the red `error_symbol` on every new shell (regression vs. the old last line, the starship guard) | Fixed in [PR #59](https://github.com/DJRHails/dotfiles/pull/59) (`ac10508b9104`): `if` form, which returns 0 on a failed guard |
| — | P4 | Missing trailing newline at EOF (pre-existing) | Fixed in the same commit |

Also verified as sound: `_cached_eval` keys the cache on argv + resolved binary path + mtime/size, so direnv upgrades/relocations invalidate correctly; `direnv hook zsh` output is static code deferring to `direnv export` at runtime, so caching it is safe; hook ordering (after starship, self-prepending to `precmd_functions`) matches direnv's documented requirement; the new lines contain no `DOTFILES_` token, so the `sed` render is unaffected.

### Verification

- **Tests**: `tests/cached-eval.test.zsh` and `tests/zellij-gc.test.zsh` — all passed
- **Syntax**: rendered template passes `zsh -n` and sources cleanly (exit 0)
- **Empirical repro**: first precmd sees `$?=1` with the merged `&&` form when direnv is absent; `$?=0` with the fix
- **Hooks**: gitleaks, trufflehog, check-crypt-patterns — all passed on commit

### Commit

`ac10508b9104` — fix(zsh): don't leak guard status 1 into first prompt (PR #58 review), on [PR #59](https://github.com/DJRHails/dotfiles/pull/59)

Verdict: approve

_[via gantry](https://gantry.hails.info/run/keen-lush-crane)_

PR #59, opened 2026-07-15
fix(zsh): don't leak direnv guard status 1 into first prompt

DJRHails:
Fixes finding F1 from the post-merge review of #58 ([inline thread](https://github.com/DJRHails/dotfiles/pull/58#discussion_r3586965391)).

**Problem:** the direnv line added in #58 is the last command in the generated zshrc. On hosts without direnv, `(( $+commands[direnv] )) && …` short-circuits with status 1; zsh hands that status to the first precmd, and starship (initialized two lines up) captures it as `STARSHIP_CMD_STATUS`, painting the red `error_symbol` on every new shell. No zlogin exists in this repo, so nothing resets the status after zshrc.

**Fix:** use the `if` form — a failed `if` with no `else` returns 0. Also adds the missing trailing newline at EOF.

**Verification:**
- Empirical repro: with the old `&&` form, first precmd sees `$?=1` when direnv is absent; with the `if` form it sees `$?=0`.
- Rendered template passes `zsh -n` and sources cleanly (exit 0).
- Both test suites pass (`tests/cached-eval.test.zsh`, `tests/zellij-gc.test.zsh`).
- Pre-commit hooks (gitleaks, trufflehog, check-crypt-patterns) all passed.

_[via gantry](https://gantry.hails.info/run/keen-lush-crane)_

### DIFF
--- a/modules/zsh/zshrc.tmpl
+++ b/modules/zsh/zshrc.tmpl
@@ -108,6 +108,9 @@ export PATH=${PATH:+$PATH:}${DOTFILES}/bin
 (( $+commands[starship] )) && _cached_eval starship init zsh
 
 # direnv: auto-load .envrc (which dotenvs .env) per directory, cached like
-# starship (see zshenv's _cached_eval). Guarded on $+commands so it no-ops
-# on hosts where direnv isn't installed.
-(( $+commands[direnv] )) && _cached_eval direnv hook zsh
+# starship (see zshenv's _cached_eval). `if` rather than `guard && cmd`
+# because this is the file's last command: a failed `&&` guard would leak
+# status 1 into the first precmd, and starship paints its prompt red on it.
+if (( $+commands[direnv] )); then
+  _cached_eval direnv hook zsh
+fi


DJRHails:
## Review Summary

Direct single-pass review (1 file, +6/−3 — follow-up delta on the already-reviewed #58, so no agent fan-out).

### Findings

No P1–P3 findings. The fix is correct and right-sized:

- In zsh, an `if` whose condition fails and has no `else` returns status 0, so the `if` form eliminates the status-1 leak from the trailing `(( $+commands[direnv] )) && …` guard.
- **Independently reproduced** on a direnv-less host (starship present): rendered the template from `origin/main` vs this branch into a throwaway `ZDOTDIR` with a first-`precmd` status probe — old form: `FIRST_PRECMD_STATUS=1`; this branch: `FIRST_PRECMD_STATUS=0`.
- **Bug-class sweep:** checked every other zsh startup file for the same trailing-guard pattern. `modules/zsh/zshenv` also ends in an `&&` guard, but zshenv runs before zshrc so its status is always overwritten before the first prompt; `zprofile` ends in `fi`; no `zlogin` exists. No siblings to fix.
- The new comment accurately states the constraint (last command in the file) — good comment, no rot.

P4 (informational, pre-existing): `modules/zsh/zshrc.local.tmpl` also lacks a trailing newline; harmless since it's sourced at the top of zshrc, left untouched.

### Verification

- **Tests**: `tests/cached-eval.test.zsh` and `tests/zellij-gc.test.zsh` both pass (all cases ok)
- **Syntax**: rendered template passes `zsh -n` for both old and new versions
- **Behavior**: first-precmd status probe confirms 1 → 0 as described above

### Commit

No fix commits needed — reviewed at `ac10508b910478440188aa242983dfae57f38c04`.

Verdict: approve

_[via gantry](https://gantry.hails.info/run/radiant-tidy-beacon)_

PR #60, opened 2026-07-20
fix(zsh): self-contained open alias, platform::open never loads

DJRHails:
`alias open='platform::open'` pointed at a function defined only in `scripts/core/platform.sh`, which the interactive startup chain never sources (`zshrc` globs `modules/*/*.zsh` only) — so every `open` died with `command not found: platform::open`, shadowing macOS's native `/usr/bin/open`.

Latent second bug: even if `platform.sh` were sourced, its macOS fallback calls `open`, which the alias re-expands to `platform::open` at definition time — infinite recursion.

Fix: replace the alias with the same idiom as the `trash-put` guard directly below it — macOS keeps native `open`; Linux gets `alias open='xdg-open'` when available. `platform::open` stays for the installer scripts that actually source it.

Verified: `zsh -ic 'type open'` → `open is /usr/bin/open`.

### DIFF
--- a/modules/zsh/aliases.zsh
+++ b/modules/zsh/aliases.zsh
@@ -6,7 +6,12 @@ alias cls='clear' # Good 'ol Clear Screen command
 
 # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 
-alias open='platform::open'
+# macOS ships /usr/bin/open natively; on Linux map `open` to xdg-open.
+# (scripts/core/platform.sh is installer-only — its platform::open never
+# exists in interactive shells, so aliasing to it broke `open` entirely.)
+if [[ "$(uname -s)" == "Linux" ]] && (( $+commands[xdg-open] )); then
+  alias open='xdg-open'
+fi
 
 if [[ "$(uname -s)" == "Linux" ]] && (( $+commands[trash-put] )); then
   alias trash='trash-put'


DJRHails:
## Review Summary

Direct single-pass review (7-line diff, one file; PR already merged, so this is a post-merge audit).

### Findings

No P1–P3 findings. The fix is correct and minimal.

| 1 | P4 | `$(uname -s)` now forks twice at shell startup (this guard + the `trash-put` guard below). `[[ $OSTYPE == linux* ]]` would avoid both forks. | Informational — the PR deliberately mirrors the existing `trash-put` idiom, and consistency wins here. Worth doing only if both guards change together. |
| 2 | P4 | On Linux without `xdg-open`, `open` is command-not-found. | Acceptable — strictly better than the previous unconditional broken alias, and matches the guarded-alias pattern. |

### Verified

- **Sourcing claim**: `zshrc.tmpl` globs only `$DOTFILES/modules/*/*.zsh` — `scripts/core/platform.sh` is indeed never sourced interactively; the old `alias open='platform::open'` could never resolve.
- **No dead code**: `platform::open` is still used by `modules/git/setup.github.sh:97` via the installer chain (`scripts/core/main.sh` sources `platform.sh`), so keeping the function is right.
- **Behavior on Linux (this container)**: without `xdg-open` the guard skips cleanly (`open` unaliased); with a stub `xdg-open` on PATH, `alias open` → `xdg-open` and invocation dispatches to it.
- **Syntax**: `zsh -n modules/zsh/aliases.zsh` clean. (shellcheck's one warning is on pre-existing line 105, outside this diff.)

### Commit

Reviewed at head `8a1c54d` — no fix commits needed.

Verdict: approve

_[via gantry](https://gantry.hails.info/run/snowy-cobalt-dune)_

PR #64, opened 2026-07-24
Durable local-session resume, resume bindings, and misc module updates

DJRHails:
Batch of accumulated working-tree changes, one logical change per commit:

- **feat(zellij): zellij::resume + cmux resume bindings** — local zellij sessions join the durable stack: `zellij::resume`, multi-host `rebuild-durable.py` (bonbon/taffy/local), per-surface cmux resume bindings for hands-free reattach after an app restart, and a TMPDIR fix for the hop husk-delete.
- **fix(zellij):** recognise `Not logged in` as a claude auth failure in durable titles.
- **chore(claude):** default model → sonnet with fast mode; `__future__` annotations in the cmux tab hook.
- **feat(pi):** `pi::glm` launcher with endpoint preflight; add `pi-web-access` extension.
- **feat(worktree):** `uv sync` Python repos on worktree bootstrap.
- **docs(skills):** notion example path fix; gantry-cli skill update (encrypted).

Note: commits are independent — prefer a rebase merge over squash to keep them distinct.


modules/agents/skills/cmux-rebuild/scripts/rebuild-durable.py
```diff
@@ -46,8 +61,32 @@
...
 )
 
+# Pane processes that do NOT make a session meaningful (zellij::sweep-husks' filter): zellij
+# itself, bare shells, and the transient helpers a prompt snapshot spawns.
+HUSK_NOISE = re.compile(
+    r"""(?x)
+      ^ (?: /[^\ ]*/ )? zellij (?: \s | $ )                                   # zellij client/server
+    | ^ (?: /usr/bin/ | /bin/ )? -? (?: zsh | bash | sh | login ) (?: \s | $ )  # bare shells
+    | snapshot-zsh .* eval                                                    # prompt snapshot
+    | ps\ -axww | ps\ -ao\ ppid | [\ /] awk\  | sed\ - | grep\  | head\ -     # transient helpers
```
DJRHails: **P2 — correctness:** the transient-helper alternatives are unanchored substrings, so real work matches as husk noise: `vim caffeinate.txt` (matches `caffeinate`), `ripgrep foo` (matches `grep\ `), `processed -x` (matches `sed\ -`), `--head -x` in any command line (matches `head\ -`). A session whose only foreground process hits one of these is judged a bare husk — it is never resumed, and worse, `idle_surfaces()` then treats its surface as safe to hijack, so `--retry` can send an attach into a live working tab: the exact failure the MEANINGFUL check exists to prevent.

Suggested fix: anchor the helpers at command start (optionally path-prefixed), the same shape as the zellij/shell alternatives:

```python
| ^ (?: /[^\ ]*/ )? (?: ps | awk | sed | grep | head | caffeinate ) (?: \s | $ )  # transient helpers
```

modules/agents/skills/cmux-rebuild/scripts/rebuild-durable.py
```diff
@@ -116,49 +261,70 @@ def list_ws() -> dict[str, str]:
...
+    barpats = [(h, re.compile(rf"Zellij \(cmux-{re.escape(h)}-([A-Za-z0-9-]+)\)")) for h in hosts]
+    meaningful = local_meaningful() if any(is_local(h) for h in hosts) else set()
+
+    def credible(h: str, m: re.Match | None) -> re.Match | None:
+        if m and is_local(h) and f"cmux-{h}-{m.group(1)}" not in meaningful:
+            return None
+        return m
+
+    out = []
     for wl in sh(["cmux", "list-workspaces"]).stdout.splitlines():
```
DJRHails: **P2 — silent failure:** `resolve_surfaces()` (here and line 333), `list_ws()`, and `idle_surfaces()` parse `sh(["cmux", ...]).stdout` without checking the exit status — bypassing the `cmux()` helper that exists to raise on failure. If the cmux CLI is failing (socket down — the post-restart state this tool is built for), stdout is empty, zero surfaces resolve, and `--retitle` prints "retitled 0 connected durable tabs" / `--bind` prints "bound 0 durable tabs" and exits 0 — a hard failure reported as a clean no-op. In `main()`, an empty `list_ws()` would also try to re-create every workspace.

Suggested fix: route the list reads (`list-workspaces`, `list-pane-surfaces`) through `cmux()` so a CLI failure aborts loudly; `read-screen` can stay best-effort.

modules/pi/aliases.zsh
```diff
@@ -19,11 +19,45 @@ export PI_SUBAGENT_SHELL_READY_DELAY_MS=3000
...
+  if [[ -z $base_url ]]; then
+    echo "pi::glm: GLM_BASE_URL is unset (checked env and $env_file)" >&2
+    echo "pi::glm: set the GLM tunnel URL in: $env_file" >&2
+    return 1
+  fi
+
+  # Probe /models before launching. A rotated trycloudflare tunnel (its DNS no
+  # longer resolves) is the usual failure; on any error point at the file to fix.
+  if ! curl -fsS --connect-timeout 3 --max-time 6 -o /dev/null \
+    -H "Authorization: Bearer $api_key" "${base_url%/}/models" 2>/dev/null; then
```
DJRHails: **P2 — misleading diagnostics:** `curl -f` fails on HTTP 401/403 exactly like on DNS/connect failure, and `2>/dev/null` discards the distinction. With a live tunnel but a wrong or empty `GLM_API_KEY` (empty sends `Authorization: Bearer ` with no warning), the user is told "the trycloudflare tunnel likely rotated — update GLM_BASE_URL", pointing them at the wrong variable.

Suggested fix: capture the HTTP code (`curl -s -o /dev/null -w '%{http_code}'`) and branch — `000` → tunnel-rotated message; `4xx/5xx` → "endpoint reachable but rejected the request (HTTP <code>) — check GLM_API_KEY".

modules/zellij/auto-attach.zsh
```diff
@@ -196,7 +220,10 @@
         # it so husks don't accumulate (otherwise every durable hop strands one detached session).
         # We're in the outer shell here (the `zellij attach` above returned on detach), so deleting
         # the session is safe. --force kills it even though its inner shell is still running.
-        { zellij delete-session --force "$session" } >/dev/null 2>&1
+        # TMPDIR must match the attach above: zellij's socket dir follows TMPDIR, and this shell's
+        # default (/var/folders/… on macOS) can't see sessions created under /tmp — the delete was
+        # silently failing and stranding every hop's husk.
+        { TMPDIR="$short_tmp" zellij delete-session --force "$session" } >/dev/null 2>&1
         _log husk-deleted "session=$session"
```
DJRHails: **P3 — log lies on failure:** line 226 suppresses the delete's exit status, then this line unconditionally logs `husk-deleted`. The comment right above documents that this exact delete silently failed for months and stranded every hop's husk — if it regresses again (another TMPDIR mismatch, a zellij CLI wedge), `attempts.log` will now actively assert success while husks accumulate, poisoning the breadcrumb used to debug it last time.

Suggested fix: branch on the delete's exit status and log `husk-deleted` vs `husk-delete-FAILED`.

modules/agents/skills/cmux-rebuild/scripts/rebuild-durable.py
```diff
@@ -1,34 +1,49 @@
...
 import re
+import socket
 import subprocess
 import sys
 import time
 from pathlib import Path
 
-HOST = os.environ.get("DURABLE_HOST") or next((a for a in sys.argv[1:] if not a.startswith("-")), "bonbon")
+LOCAL_HOST = socket.gethostname().split(".")[0].lower()
+_ARG_HOSTS = [a for a in sys.argv[1:] if not a.startswith("-")]
```
DJRHails, line 38: **P3 — no unknown-flag guard on a mutating tool:** hand-rolled `in sys.argv` parsing silently drops mistyped flags — `--dry_run`, `--dryrun`, or `-n` are ignored and the script proceeds with the real pass (deleting `wrap-` sessions, creating workspaces, sending attaches). Relatedly, `--retry`/`--retitle`/`--bind` silently ignore `--dry-run` — `rebuild-durable.py --retry --dry-run` sends real attaches.

Suggested fix: reject unknown `-`-prefixed args up front, and reject `--dry-run` combined with the mode flags (or honor it there).

modules/zellij/durable-remote.sh
```diff
@@ -291,7 +291,7 @@ for s in $sessions; do stale "$cdir/$s" && { need_summaries=1; break; }; done
 
 if [ "$need_summaries" = 1 ] && ! probe=$(auth_check); then
   case "$probe" in
-    *401* | *authenticate* | *'Invalid authentication'*)
+    *401* | *authenticate* | *'Invalid authentication'* | *'Not logged in'*)
```
DJRHails: **P3 — same bug class as this commit's fix:** claude's other common auth phrasing, `Invalid API key · Please run /login`, matches none of these patterns (no `401`, no lowercase `authenticate`, not `Invalid authentication` / `Not logged in`) — so the probe result falls through, `need_summaries` stays 1, every session runs a doomed `summarise`, and the picker shows `?` titles with no `__AUTHFAIL__` diagnosis: exactly the silence the `Not logged in` addition is fixing. Same gap in the `summarise` case at line 129.

Suggested fix: add `*'Invalid API key'*` to both case patterns.

modules/agents/skills/worktree/scripts/worktree-create.sh
```diff
@@ -316,6 +316,18 @@ worktree::bootstrap() {
     direnv allow "$wt" 2>/dev/null || true
   fi
 
+  # Python/uv repos: install the project venv so the worktree is runnable, not just checked out —
+  # the Python analogue of the node_modules CoW clone above (a fresh worktree has no .venv, and .venv
+  # lives at the repo root, not under the symlinked .data, so it is not shared). uv's global cache
+  # hardlinks packages, so this is fast. Base deps only; a repo's dev extra is installed on demand.
```
DJRHails: **P3 — comment inaccurate:** bare `uv sync` installs the default dependency groups, which **includes the `dev` group** by default (`--no-dev` excludes it). "Base deps only; a repo's dev extra is installed on demand" is only true for repos that model dev deps as a PEP-621 extra. Either add `--no-dev` (if base-only is the intent) or fix the comment to "default groups (incl. `dev`); extras are installed on demand".

modules/agents/skills/cmux-rebuild/SKILL.md
```diff
@@ -28,39 +30,92 @@ fresh. Non-interactive: `--list`, `--attach <session>`, `--query <str>`, `--reap
...
+python3 scripts/rebuild-durable.py [hosts...] --bind      # backfill cmux resume bindings (see below)
 ```
 
-It sorts each live remote session into a cmux workspace named after its repo (cwd basename), creates
-a surface, and `cmux send`s `ssh::durable <host> --attach <session>` (the current de-nesting attach).
-Idempotent — already-connected sessions are skipped (port check), so re-running is safe.
+It sorts each live session into a cmux workspace named after its repo (cwd basename), grabs a
+surface (the workspace's initial one first, then `new-surface`), and `cmux send`s the attach —
+`ssh::durable <host> --attach <session>` for remote hosts, `zellij::resume <session>` for local
+ones. Idempotent — connected sessions are skipped (remote: exact mosh-port check; local: a live
```
DJRHails: **P3 — doc overpromise:** "Idempotent — connected sessions are skipped … so re-running is safe" only holds once attaches have settled. For pre-existing workspaces the default pass always mints a fresh `new-surface` for every session still in `connect` status (the `spare` reuse only covers workspaces the script itself just created), so re-running mid-settle mints duplicate surfaces — producing the very "Idle leftover surfaces" gotcha documented below. `--retry` is the actually-idempotent path.

Suggested wording: "Idempotent once connections settle — … mid-settle, use `--retry` (which reuses surfaces) rather than re-running the default pass."

DJRHails (commented):
Automated review — findings posted inline. Each thread is resolved as its fix lands.

modules/agents/skills/cmux-rebuild/scripts/rebuild-durable.py
DJRHails: Fixed in 629784b82d63 — helpers anchored at command start (`^(?:/[^ ]*/)?(?:ps|awk|sed|grep|head|caffeinate)(?:\s|$)`). Same fix applied to the sibling filter in `zellij::sweep-husks` (mosh-zellij.zsh:442), and a table-driven regression test pins both directions (tests/rebuild_durable_test.py).

modules/agents/skills/cmux-rebuild/scripts/rebuild-durable.py
DJRHails: Fixed in 629784b82d63 — `list-workspaces`/`list-pane-surfaces` reads in `list_ws()`, `idle_surfaces()`, `resolve_surfaces()`, and `main()`'s spare-surface scan now go through `cmux()`, which raises on a nonzero exit. `read-screen` left best-effort as suggested.

modules/pi/aliases.zsh
DJRHails: Fixed in 629784b82d63 — the preflight now captures `%{http_code}`: `000`/empty keeps the tunnel-rotated message, any HTTP rejection reports the code and points at GLM_API_KEY.

modules/zellij/auto-attach.zsh
DJRHails: Fixed in 629784b82d63 — the delete now branches on its exit status and logs `husk-deleted` vs `husk-delete-FAILED`.

modules/agents/skills/cmux-rebuild/scripts/rebuild-durable.py
DJRHails: Fixed in 629784b82d63 — `main()` rejects unknown `-`-prefixed args and the `--dry-run` + mode-flag combination up front (validated in `main()` rather than at import so the module stays importable for tests). Exercised both rejection paths by hand.

modules/zellij/durable-remote.sh
DJRHails: Fixed in 629784b82d63 — `*'Invalid API key'*` added to both case patterns (summarise and the auth_check probe).

modules/agents/skills/worktree/scripts/worktree-create.sh
DJRHails: Fixed in 629784b82d63 — comment corrected to "Default groups only (uv includes the dev group by default); extras are installed on demand." Kept the behavior (dev group installed): worktree agents usually want pytest et al., so `--no-dev` would be the wrong default.

modules/agents/skills/cmux-rebuild/SKILL.md
DJRHails: Fixed in 629784b82d63 — reworded to "Idempotent once connections settle" and pointed mid-settle re-runs at `--retry`.

DJRHails:
## Review Summary

Multi-agent review (five parallel pr-review-toolkit passes: code quality, silent failures, comment accuracy, test coverage, type design) — 669 added lines of real source logic warranted the full battery. The `gantry-cli/SKILL.md` hunk is transcrypt ciphertext and was not reviewed. All 8 P1–P3 findings were fixed in [`629784b82d63`](https://github.com/DJRHails/dotfiles/commit/629784b82d63); every inline thread is resolved with a pointer to its fix.

### Findings

| F1 | P2 | `HUSK_NOISE` transient-helper patterns were unanchored substrings — real work (`vim caffeinate.txt`, `ripgrep foo`) classified as husk noise, so `--retry` could hijack a live tab | Fixed: anchored at command start; same fix in the sibling `zellij::sweep-husks` awk filter; regression test added (`tests/rebuild_durable_test.py`) |
| F2 | P2 | `list_ws()`/`idle_surfaces()`/`resolve_surfaces()` ignored the cmux exit status — a down cmux CLI made `--retitle`/`--bind` report "0 tabs" and exit 0 | Fixed: list reads routed through the raising `cmux()` helper; `read-screen` stays best-effort |
| F3 | P2 | `pi::glm` preflight misdiagnosed HTTP auth rejection as a rotated tunnel | Fixed: branches on `%{http_code}` — 000 → tunnel message, HTTP error → points at `GLM_API_KEY` |
| F4 | P3 | `_log husk-deleted` asserted success even when the delete failed (the exact silent failure this PR documents) | Fixed: logs `husk-deleted` vs `husk-delete-FAILED` on the actual exit status |
| F5 | P3 | Mistyped flags silently ran the mutating pass; `--retry --dry-run` sent real attaches | Fixed: `main()` rejects unknown flags and the `--dry-run` + mode-flag combination |
| F6 | P3 | Auth patterns miss claude's `Invalid API key · Please run /login` phrasing — same bug class as this PR's `Not logged in` fix | Fixed: added `*'Invalid API key'*` to both case patterns |
| F7 | P3 | "Base deps only" comment wrong — bare `uv sync` includes the `dev` group | Fixed: comment corrected; behavior kept (worktree agents want dev deps) |
| F8 | P3 | SKILL.md idempotence claim — re-running the default pass mid-settle mints duplicate surfaces | Fixed: scoped to "once connections settle", mid-settle re-runs pointed at `--retry` |

P4 (informational, not changed): the SKILL.md quick-ref now says `cmux workspace list` while the script and three sibling skill docs use `cmux list-workspaces` — if both are valid aliases fine, otherwise worth reconciling; default `HOSTS` isn't deduped if the script ever runs on bonbon/taffy itself; the two zellij calls in `main()`'s wrap-stub cleanup have no `timeout=`; `zellij::resume`/`ssh::durable::attach` set resume bindings silently (no `set-FAILED` breadcrumb, unlike auto-attach.zsh) and unconditionally; `DURABLE_HOST` overriding the default host list is undocumented in SKILL.md; `local_cwd()` timeout silently files a session under `home`.

### Verification

- **Tests**: new `tests/rebuild_durable_test.py` passes (`python3 tests/rebuild_durable_test.py` — pytest not installed in this sandbox); existing `tests/zellij-gc.test.zsh` all 10 checks pass
- **Lint**: `ruff check` clean (both Python files + test); `shellcheck` clean on changed shell (SC2015 at `worktree-create.sh:230` is pre-existing, outside this diff)
- **Syntax**: `py_compile` clean; `zsh -n` clean on all three changed zsh files; both settings JSONs valid
- **Behavior**: flag-guard rejection paths exercised by hand; fixed awk/Python husk filters verified against a classification table
- Pre-commit hooks (gitleaks, trufflehog, transcrypt-pattern check) passed on commit; code-simplifier polish pass made no changes

### Commit

`629784b82d63` — fix: resolve code review findings for PR #64

Verdict: approve

_[via gantry](https://gantry.hails.info/run/valiant-cool-moor)_

PR #65, opened 2026-07-24
Migrate at-rest encryption from transcrypt to glassine

DJRHails:
Moves the repo's at-rest encryption from transcrypt (openssl aes-256-cbc, shared password) to [glassine](https://github.com/DJRHails/glassine) v0.6.0 (sops envelopes, per-file data keys wrapped to the SSH keys published at github.com/DJRHails.keys, recipients tracked in `.sops.yaml`).

- **Same boundary**: identical globs, same 188 files (set-diffed old vs new attribute matches — no strays). History is not rewritten; pre-migration blobs stay transcrypt ciphertext, with the recovery recipe in `docs/legacy-transcrypt.md` (password placeholder — fill locally, file is itself encrypted).
- **Ported integrations**: `check-crypt-patterns.sh`, the encryption-boundary Claude hook, the zshrc undecrypted-module skip (now detects sops envelopes instead of the OpenSSL `Salted__` magic — verified against a real envelope), and `modules/git/setup.sh` now installs sops + glassine.
- **Enforcement actually wired**: global `core.hooksPath` (~/.git-hooks lefthook shim) has been bypassing `.git/hooks/pre-commit` — the transcrypt vendor hook was silently dead. New `lefthook.yml` runs `check-crypt-patterns.sh` + `glassine check` on every commit (verified live), and `.pre-commit-config.yaml` gains a matching `glassine check` entry for prek.
- **`.dockerignore` drift fixed**: five encrypted paths were missing, so their decrypted plaintext entered the docker build context.
- **Per-clone cutover**: `docs/glassine-migration.md` — one command per host after pulling (installs sops/glassine if missing, `glassine init`, `transcrypt --uninstall`, verifies every managed file is an envelope). Also covers the checkout-across-the-boundary re-smudge gotcha.

Verified locally: all 188 files encrypt in the index / stay plaintext in the worktree, round-trip decrypt works (identity auto-pinned), pattern + encryption checks green.


docs/glassine-migration.md
```diff
@@ -0,0 +1,54 @@
...
+
+```bash
+cd ~/.files \
+  && command -v sops >/dev/null || brew install sops \
+  ; command -v glassine >/dev/null || { mkdir -p ~/.local/bin \
+      && curl -fsSL https://raw.githubusercontent.com/DJRHails/glassine/main/glassine \
+         -o ~/.local/bin/glassine && chmod +x ~/.local/bin/glassine; } \
+  ; glassine init \
+  && transcrypt --uninstall --yes 2>/dev/null \
+  ; glassine status | awk '$1 != "encrypted"' | grep . && echo 'CHECK FAILED' || echo OK
```
DJRHails: **P2 — fail-open verification:** if `glassine status` itself fails (glassine or sops not installed, `init` failed), the pipe is empty, `grep .` exits non-zero, and the runbook prints `OK` — exactly the hosts most likely to be broken get a false pass. Also `awk '$1 != "encrypted"'` flags the `empty` state, which glassine treats as a by-design pass-through, so a legitimately empty managed file would print `CHECK FAILED`.

Suggested: inject a sentinel line on failure so it lands in the awk filter, and accept `empty`:

```bash
; { glassine status || echo 'status-failed'; } | awk '$1 !~ /^(encrypted|empty)$/' | grep . && echo 'CHECK FAILED' || echo OK
```

docs/glassine-migration.md
```diff
@@ -0,0 +1,54 @@
...
+History was NOT rewritten: pre-migration blobs remain transcrypt ciphertext
+(readable per docs/legacy-transcrypt.md).
+
+## Per-clone cutover (every host with a checkout: trifle, bonbon, taffy, …)
+
+After pulling the migration commit, run once per clone:
+
+```bash
+cd ~/.files \
+  && command -v sops >/dev/null || brew install sops \
```
DJRHails: **P3 — brew-only sops install on Linux targets:** the runbook explicitly targets bonbon and taffy (Linux), but `brew install sops` only works where (linux)brew exists — on a bare host it fails and cascades into the fail-open check above. `modules/git/setup.sh` already solves this with a GitHub-release fallback; the one-liner should degrade the same way (current stable is v3.13.3, assets `sops-v<ver>.linux.{amd64,arm64}`).

DJRHails (commented):
Automated review — findings posted inline. Each thread is resolved as its fix lands. Boundary claims were verified empirically: all 189 glassine-managed blobs in HEAD are genuine sops envelopes (zero plaintext), the managed set is exactly the old 188 transcrypt files plus `docs/legacy-transcrypt.md`, every envelope wraps to the six `.sops.yaml` recipients (no drift), and every managed path is excluded by `.dockerignore`.

docs/glassine-migration.md
DJRHails: Fixed in c48e20d557dd — a failing `glassine status` now injects a sentinel line that trips the check (verified by simulation: missing glassine → CHECK FAILED), and the awk filter accepts the by-design `empty` state.

docs/glassine-migration.md
DJRHails: Fixed in c48e20d557dd — added a GitHub release-binary fallback (sops v3.13.3, amd64/arm64 via `uname -m`) mirroring `modules/git/setup.sh`; brew stays the preferred path where present. Release URL verified (HTTP 200).

DJRHails:
## Review Summary

Direct single-pass review (13 hand-written files, ~138 lines; the other 189 changed files are glassine re-encryptions). Because this PR moves the encryption boundary itself, the PR's security claims were verified empirically rather than taken from the description:

- **All 189 glassine-managed blobs in HEAD are genuine sops envelopes** — zero plaintext committed (checked every blob via `git cat-file`).
- **Boundary parity**: the `filter=glassine` set is exactly the old 188 `filter=crypt` files plus the new `docs/legacy-transcrypt.md` (set-diffed via `git check-attr --source`).
- **No recipient drift**: every envelope wraps to precisely the six `.sops.yaml` keys (jq over all 189 `.sops.age[].recipient` sets).
- **`.dockerignore` covers every managed path** (per-path prefix check).
- **zshrc envelope detection** (`{` + `ENC[AES256_GCM`) matches the first two lines of all 189 real envelopes.
- `install::release_binary` prefers brew, so the Linux-only sops URL in `modules/git/setup.sh` is only the bare-Linux fallback — correct.

### Findings

| F1 | P2 | Cutover runbook verification was fail-open: a failing `glassine status` (glassine/sops missing) printed `OK`; and `empty` — a by-design pass-through state — tripped `CHECK FAILED` | Fixed: sentinel line on status failure + accept `empty` ([thread](https://github.com/DJRHails/dotfiles/pull/65#discussion_r3642553520)) |
| F2 | P3 | Runbook installed sops via brew only, but explicitly targets brew-less Linux hosts (bonbon, taffy) | Fixed: GitHub release-binary fallback (v3.13.3), mirroring `modules/git/setup.sh` ([thread](https://github.com/DJRHails/dotfiles/pull/65#discussion_r3642553522)) |
| F3 | P4 | `setup.sh` installs glassine unpinned from `main` | Noted only — deliberate (single script, no release artifacts, own repo) and documented in the code comment |
| F4 | P4 | Binary assets under glassine (diagram-design PNGs/WOFFs) will emit a non-fatal "BINARY under glassine" advisory from `glassine check` on every commit touching them | Noted only — pre-existing boundary, advisory is informational by design |

### Verification

- **Encryption checks**: `scripts/check-crypt-patterns.sh` and `glassine check` green (run locally with glassine 0.6.0)
- **Runbook one-liner**: `bash -n` clean; simulated missing-glassine / all-encrypted+empty / plaintext cases produce `CHECK FAILED` / `OK` / `CHECK FAILED`; sops release URL returns HTTP 200
- **Lint**: shellcheck on the three changed scripts — only pre-existing info/warnings outside the diff
- **Hooks at commit time**: gitleaks, TruffleHog, check-crypt-patterns, glassine-check all passed

### Commit

c48e20d557dd `fix: resolve code review findings for PR #65`

Verdict: approve

_[via gantry](https://gantry.hails.info/run/amber-snowy-meadow)_

DJRHails (commented):
Delta re-review of 85e07978c454 (scoped read access for NxtGenLegend). Verified empirically: the five `github:NxtGenLegend` keys in `.sops.yaml` exactly match github.com/NxtGenLegend.keys; all 61 diagram-design envelopes carry precisely the scoped 11-key set; the other 128 envelopes still carry only the six DJRHails keys (no leakage outside the scoped path); rule ordering is correct (scoped rule before the catch-all, sops first-match). One new finding inline.

.sops.yaml
```diff
@@ -0,0 +1,24 @@
+# managed by glassine — edit recipients via `glassine allow` / `glassine revoke`
+creation_rules:
+  - path_regex: '^modules/agents/skills/diagram-design/.*$' # glob: modules/agents/skills/diagram-design/**
```
DJRHails, line 3: **P2 — glassine version skew breaks commits on already-cutover hosts:** scoped `path_regex` rules are only understood by glassine ≥ 0.7.0 (released today; the migration shipped against 0.6.0). Verified: `glassine check` at this head is green under 0.7.0 but under 0.6.0 false-flags all ~128 catch-all-rule envelopes as `RECIPIENT DRIFT` and dies — and lefthook, prek, and the runbook's `.git/hooks/pre-commit` chain all run it, so every commit fails. Nothing upgrades an existing install: both `modules/git/setup.sh` and the runbook one-liner skip when `command -v glassine` succeeds, so any host that cut over before today keeps 0.6.0 indefinitely. The 0.6.0 error text ("run `glassine rotate` and commit") points at the wrong fix.

Suggested: make the runbook's glassine install an unconditional refresh (it's a single idempotent script fetch) and add a note that mass RECIPIENT DRIFT after this commit means *upgrade glassine*, not rotate.

DJRHails: Fixed in 1731c5e454c9 — the cutover one-liner now refreshes glassine from main unconditionally, and a note explains that mass RECIPIENT DRIFT means a stale glassine (<0.7.0), not real drift — upgrade, do not rotate. Verified `glassine check` green under 0.7.0 at this head.

DJRHails:
## Delta Review Summary (85e07978c454 — scoped read access for NxtGenLegend)

Direct delta re-review: one hand-written file changed (`.sops.yaml`, +14 lines); the other 61 files are diagram-design envelope rotations. The grant's blast radius was verified empirically:

- **The five `github:NxtGenLegend` keys exactly match github.com/NxtGenLegend.keys** (bidirectional diff).
- **Scoping holds**: all 61 diagram-design envelopes carry precisely the scoped 11-key set (5 NxtGenLegend + 6 DJRHails); the other 128 envelopes still carry only the six DJRHails keys — nothing outside `modules/agents/skills/diagram-design/` is readable by the new keys.
- **Rule ordering is correct**: the scoped `path_regex` rule precedes the catch-all, and sops uses first-match.

### Findings

| F5 | P2 | glassine version skew: scoped `path_regex` rules need glassine ≥ 0.7.0 (released today). Hosts that cut over earlier keep 0.6.0 forever (both `setup.sh` and the runbook skipped install when present); 0.6.0's `check` flattens the policy, false-flags all ~128 catch-all envelopes as RECIPIENT DRIFT, and blocks every commit — with an error suggesting `glassine rotate`, which cannot fix it | Fixed: runbook now refreshes glassine unconditionally + a note that mass RECIPIENT DRIFT means upgrade, not rotate ([thread](https://github.com/DJRHails/dotfiles/pull/65#discussion_r3647570477)) |

### Verification

- `glassine check` + `scripts/check-crypt-patterns.sh`: green at this head under glassine 0.7.0 (and the 0.6.0 failure mode reproduced live to confirm the finding)
- Runbook one-liner: `bash -n` clean
- Commit hooks (gitleaks, TruffleHog, crypt-patterns, glassine-check): all passed

### Commit

1731c5e454c9 `fix: resolve delta review finding for PR #65`

Verdict: approve

_[via gantry](https://gantry.hails.info/run/amber-snowy-meadow)_

PR #67, opened 2026-07-24
Add per-agent gantry worker keys as glassine recipients

DJRHails:
Six unique ed25519 keys — one per dotfiles-provisioned gantry agent, generated on the fleet host at `~/.gantry/agent-keys/<agent>/` — become recipients, replacing the shared transcrypt password the worker images baked in at build time. Repo-wide scope per explicit owner decision (parity with the previous shared password; derived images inherit the base image's decrypted files regardless). Each agent is now individually revocable: `glassine revoke 'gantry:<agent>'`.

Also fixes `check-crypt-patterns.sh` to skip all of `.data/` (a leftover worktree's .gitattributes read as 13 dead rules and blocked commits).

Companion gantry-repo PR (image bootstrap → glassine) to follow.


DJRHails:
## Review Summary

Direct single-pass review (hand-written diff is 2 files; the other 196 changed files are glassine ciphertext churn from re-encryption). **No P1–P3 findings** — no inline threads, no fix commit needed.

### Findings

| 1 | P4 | The 6-key gantry recipient block is duplicated verbatim across both creation rules in `.sops.yaml`. A YAML anchor could dedupe it, but only if glassine's parser supports anchors — not worth the risk/churn for 2 rules. | Informational |
| 2 | P4 | `shfmt -i 2` still flags two pre-existing hunks in `check-crypt-patterns.sh` (case-arm indent, `done <"$attr"`) — lines this PR didn't touch. | Informational |

### Verification

- **Re-encryption completeness**: all 196 glassine-filtered tracked files (exact set — none missed) parsed at HEAD; every one carries all 6 new `gantry:*` recipients and **zero pre-existing recipients were dropped** (checked programmatically against `origin/main`). Recipient counts match the rules: 12 on default-rule files, 17 on diagram-design files.
- **Key validity**: all 6 added keys base64-decode to well-formed `ssh-ed25519` recipients (32-byte points), all distinct, no collision with existing recipients.
- **Script fix**: `scripts/check-crypt-patterns.sh` runs clean at HEAD (exit 0 — the 13 dead-rule false positives from the stray `.data/` worktree copy are gone); `shellcheck` clean.
- **Hooks**: `glassine check` clean; script is wired identically in both `lefthook.yml` and `.pre-commit-config.yaml` (unchanged).
- **Not verified**: end-to-end decryption with an actual gantry agent private key — this sandbox has no key access (denied by policy). The structural recipient check above is the evidence; worth a one-time `glassine` smoke test from a worker once the companion gantry-repo bootstrap PR lands.

### Commit

No fix commit — PR head `0b58adc` reviewed as-is.

Verdict: approve

_[via gantry](https://gantry.hails.info/run/ancient-velvet-leaf)_

PR #69, opened 2026-07-25
pi: pin subagents to claude-opus-5[fast]

DJRHails:
All five pi subagent definitions still pinned `claude-opus-4-8[fast]`. Bump them to `claude-opus-5[fast]`, matching the default already set in `modules/pi/settings.json`.

Verified against `pi models`: `claude-opus-5[fast]` is registered on the `anthropic-api` provider (1M ctx / 128K out), so no ambiguous-id routing risk.

Note: `~/.pi/agent/agents/*.md` on this host are plain copies rather than the symlinks declared in `modules/pi/symlinks.conf` — they were updated in lockstep by hand. A `scripts/link.sh` run would re-establish the links.

### DIFF
--- a/modules/agents/AGENTS.md
+++ b/modules/agents/AGENTS.md
@@ -175,7 +175,7 @@ Pin actions to version tags: `actions/checkout@v4` (use `persist-credentials: fa
   fit the task — no Opus pin.** `anthropic-api` uses the load-balanced `ANTHROPIC_API_KEY_*` keys
   and works wherever they resolve. **Any registered `anthropic-api` model works, including non-Opus
   (`claude-fable-5`, `claude-haiku-4-5`, `claude-sonnet-5`)** — use a cheap/small model for recon,
-  reviews, and parallel fan-out; reserve `claude-opus-4-8[fast]` for work that needs the frontier.
+  reviews, and parallel fan-out; reserve `claude-opus-5[fast]` for work that needs the frontier.
   The old "always pin to `claude-opus-4-8[fast]`" rule was a *workaround* for a real bug, fixed by
   `DJRHails/pi-interactive-subagents` PRs #9 and #12 (still open as of 2026-07-24, not yet on
   `main`): a bare/ambiguous model id (e.g. `claude-fable-5`) fell through to pi's keyless built-in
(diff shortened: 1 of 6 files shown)


DJRHails:
## Review Summary

Direct single-pass review (+5/−5 frontmatter-only diff across 5 files — below the multi-agent threshold). The change itself is correct and verified; one stale-guidance sibling was fixed alongside it.

### Findings

| 1 | P3 | `modules/agents/AGENTS.md` still recommended reserving `claude-opus-4-8[fast]` for frontier work — the same stale-pin class this PR removes | Fixed in c4f2bf556857: live guidance now says `claude-opus-5[fast]`; the historical bug-workaround narrative below it intentionally keeps the old id |
| 2 | P4 | `pi::ant` / `pi::ant-batch` aliases (`modules/pi/aliases.zsh`) still pin `claude-opus-4-8` variants | Not changed: they target the `ant-fellow-*` gateway, a different provider whose model availability can't be verified from this sandbox — deliberate, out of scope |
| 3 | P4 | `modules/claude/hooks/experiment-log.sh` `CURATE_MODEL` defaults to `claude-opus-4-8` | Not changed: a direct-API model id (not a pi virtual pin) and still valid |

### Verification

- **Pin correctness**: `claude-opus-5[fast]` is registered on the `anthropic-api` provider (`modules/pi/extensions/anthropic-api-provider.ts:202`; confirmed live via `pi models` in the review container) and matches the `defaultModel` in `modules/pi/settings.json`.
- **No ambiguous-id routing risk**: the `[fast]` virtual ids exist only in the `anthropic-api` extension — `anthropic-oauth-providers.ts` is a no-op stub — so the pin keeps the uniqueness property the old workaround relied on while pi-interactive-subagents PRs #9/#12 remain unmerged.
- **Frontmatter**: all five agent files parse and carry the identical pin (scripted check).
- **Hooks**: gitleaks, TruffleHog, and both glassine checks passed on commit.
- **Tests**: repo test suite (`tests/`) does not cover the pi module — nothing applicable to run. No CI workflows exist.

### Commit

c4f2bf556857 — `fix: resolve code review findings for PR #69`

Verdict: approve

_[via gantry](https://gantry.hails.info/run/fair-ember-owl)_

PR #72, opened 2026-07-26
fix(git): scope the GitHub https->ssh rewrite to hosts with an ssh key

DJRHails:
## Problem

The tracked `modules/git/gitconfig` sets `url.git@github.com:.insteadOf = https://github.com/` globally. On gantry djrhails-dev workers (no GitHub ssh key; token-over-https auth) it collides with the worker image's system-scope `url.https://github.com/.insteadOf` rules.

Core git survives the collision via a deterministic tie-break (first-longest wins, system read first — the image's identity mapping relies on this). **git-lfs does not**: `lfsapi/endpoint_finder.go` builds its alias map by iterating a Go map with last-write-wins, so with two config keys claiming the alias `https://github.com/` the winner is **random per invocation**. Result: ~50% of pushes in LFS-enabled repos (e.g. touchstone) die at the LFS pre-push hook with `Host key verification failed.: exit status 255`. Reported by gantry run [mild-clever-thistle](https://gantry.hails.info/run/mild-clever-thistle) — its log shows an identical push failing then succeeding seconds apart. Reproduced on a fresh worker: 10/20 dry-run pushes failed.

`insteadOf` rules cannot be unset by a later config file, so the rewrite has to leave the tracked global config.

## Fix

- `modules/git/gitconfig`: the https→ssh rewrite is replaced by an optional include of `~/.gitconfig.github-ssh` (git silently skips a missing include, so keyless hosts stay on https).
- `modules/git/setup.sh`: generates that file only when `~/.ssh/config` has the `Host github.com` block that `github::set_ssh_key` writes; re-checks after an interactive `github::setup` provisions a new key. Workers never have that block (gantry setup only writes cluster host blocks).
- `modules/dotfiles-autoupdate/update.sh`: ensures the file daily on already-provisioned hosts, so existing personal machines regain the rewrite on their next autoupdate without a manual setup run. (Gap: the ensure-step ships in the same pull that drops the rule, so a personal host runs ~one day on https+gh-credential-helper before the next timer run migrates it.)

## Verification

- Fixed config on a worker: LFS endpoint resolves https deterministically, duplicate-alias warning gone, **20/20 clean pushes** (was 10/20 failing).
- Simulated personal host (`Host github.com` in ssh config): setup generates the include and `git ls-remote --get-url https://github.com/DJRHails/dotfiles.git` → `git@github.com:DJRHails/dotfiles.git` (behavior preserved).
- `shellcheck` clean on the edited scripts (pre-existing infos unchanged).

Companion PR adds a duplicate-alias build guard to the gantry worker image.

_[via gantry](https://gantry.hails.info/run/stark-dusk-cairn)_


modules/git/setup.sh
```diff
@@ -32,10 +32,33 @@ setup_gitconfig () {
...
 
+setup_github_ssh_rewrite () {
+  # The tracked gitconfig includes ~/.gitconfig.github-ssh (missing file =
+  # silently skipped). Generate it only where the GitHub SSH key is set up —
+  # github::set_ssh_key writes the "Host github.com" block this checks for.
+  # Headless workers (gantry) never have that block and must stay on HTTPS:
+  # a global https->ssh rewrite there collides with their ssh->https rule and
+  # git-lfs resolves the duplicate alias nondeterministically, killing ~50%
+  # of pushes in LFS repos.
+  if grep -qs '^Host github.com$' "$HOME/.ssh/config"; then
```
DJRHails: **P3 — robustness:** `grep -qs '^Host github.com$'` only matches the exact line `github::add_ssh_configs` writes. ssh_config legitimately allows indentation, a lowercase `host` keyword, trailing whitespace, and multiple patterns per line (`Host github.com gist.github.com` is a common hand-written stanza). A host with any of those variants *had* the rewrite before this PR (it was unconditional) and silently loses it after — the failure mode is the original one this comment block describes: https submodules breaking on Linux. A tolerant-but-anchored match keeps worker configs, `*.github.com.internal` lookalikes, commented lines, and `!github.com` negations excluded:

```sh
grep -qsiE '^[[:space:]]*Host[[:space:]]+([^#]*[[:space:]])?github\.com([[:space:]]|$)'
```

Same pattern needs to land in `ensure_github_ssh_rewrite` in `modules/dotfiles-autoupdate/update.sh` — the two checks must stay in sync.

modules/dotfiles-autoupdate/update.sh
```diff
@@ -100,5 +100,28 @@ refresh_pi_packages() {
...
+# a key (it writes the "Host github.com" block), and must not exist on
+# headless workers, which push over HTTPS with a token. Running this daily
+# migrates hosts that were provisioned before the rewrite moved out of the
+# tracked gitconfig, without waiting for a manual setup run.
+ensure_github_ssh_rewrite() {
+  local rewrite="$HOME/.gitconfig.github-ssh"
+  if grep -qs '^Host github.com$' "$HOME/.ssh/config"; then
+    if [ ! -f "$rewrite" ]; then
+      printf '%s\n' \
+        '# Generated by modules/git/setup.sh: this host has an authorised GitHub' \
```
DJRHails: **P3 — provenance:** when the autoupdate creates this file, its header claims `# Generated by modules/git/setup.sh` — someone debugging a surprise `~/.gitconfig.github-ssh` on a host that never ran setup will look in the wrong script and miss the daily ensure-step entirely. Use one neutral header naming both generators, byte-identical in both scripts so the generated file never reveals (or misstates) which path created it.

modules/dotfiles-autoupdate/update.sh
```diff
@@ -100,5 +100,28 @@ refresh_pi_packages() {
...
+# tracked gitconfig only *includes* that file (see modules/git/gitconfig): the
+# HTTPS -> SSH rewrite belongs on hosts where github::set_ssh_key provisioned
+# a key (it writes the "Host github.com" block), and must not exist on
+# headless workers, which push over HTTPS with a token. Running this daily
+# migrates hosts that were provisioned before the rewrite moved out of the
+# tracked gitconfig, without waiting for a manual setup run.
+ensure_github_ssh_rewrite() {
+  local rewrite="$HOME/.gitconfig.github-ssh"
+  if grep -qs '^Host github.com$' "$HOME/.ssh/config"; then
+    if [ ! -f "$rewrite" ]; then
```
DJRHails: **P3 — stale content:** the `[ ! -f "$rewrite" ]` guard means this function only ever *creates* the file; if the generated content changes in a future commit (even a comment fix), hosts that already have the file keep the old content forever, contradicting the function comment's "Keep ~/.gitconfig.github-ssh in step". Compare desired content against the existing file and rewrite on mismatch — still idempotent (writes and logs only on change), and the autoupdate then genuinely converges drift.

DJRHails (commented):
Automated review — findings posted inline. Each thread is resolved as its fix lands. Core fix verified sound: with the include absent a worker resolves LFS/https URLs deterministically (no duplicate alias key anywhere), and with the `Host github.com` block present the include rewrites `https://github.com/` → `git@github.com:` exactly as before.

modules/git/setup.sh
DJRHails: Fixed in fe4698de447f.

modules/dotfiles-autoupdate/update.sh
DJRHails: Fixed in fe4698de447f.

modules/dotfiles-autoupdate/update.sh
DJRHails: Fixed in fe4698de447f.

DJRHails:
## Review Summary

Direct single-pass review (64-line, 3-file shell/config change — below the multi-agent threshold). The core fix is sound and was verified behaviorally, not just by reading: with the include file absent, a worker resolves `https://github.com/` URLs deterministically and no `url.git@github.com:.insteadOf` key exists in any scope (the git-lfs duplicate-alias collision is gone); with a `Host github.com` entry present, the include restores the exact pre-PR https→ssh rewrite (`git ls-remote --get-url` → `git@github.com:…`).

### Findings

| F1 | P3 | `grep -qs '^Host github.com$'` matched only the exact provisioned line — hosts with legitimate ssh_config variants (indentation, lowercase `host`, `Host github.com gist.github.com`, trailing whitespace) silently lost the rewrite they had before this PR | Fixed: tolerant-but-anchored match in both scripts, validated against 9 config variants (matches all legit forms; still rejects worker configs, `*.github.com.internal` lookalikes, commented lines, `!github.com` negations) |
| F2 | P3 | Autoupdate-generated file claimed `# Generated by modules/git/setup.sh` — wrong provenance for anyone debugging a surprise `~/.gitconfig.github-ssh` | Fixed: one neutral header naming both generators, byte-identical in both scripts (verified with `cmp`) |
| F3 | P3 | `ensure_github_ssh_rewrite` only created the file when missing, so future content changes would never reach already-migrated hosts | Fixed: compares desired content and rewrites on mismatch — converges a stale file once, then stays idle |
| — | P4 | The ensure-step converges in one direction only: it never removes the file if a host loses its GitHub key. Deliberate — auto-deleting config is riskier than leaving it; noted for awareness | Informational |
| — | P4 | The acknowledged one-day gap (a personal host runs on https + gh helper between pulling this change and its next autoupdate) is fine as described in the PR body | Informational |

### Verification

- **Behavior**: sandboxed-`$HOME` end-to-end tests — worker config creates nothing; provisioned and hand-written-variant hosts both get byte-identical files from either generator; stale content converges then goes idempotent; https→ssh rewrite verified via `git ls-remote --get-url` with `GIT_CONFIG_NOSYSTEM=1`
- **Lint**: `shellcheck` — no new findings (pre-existing SC1091 infos / SC2154 on untouched lines unchanged)
- **Secrets**: `gitleaks` staged scan + `trufflehog` filesystem scan clean (golang-based prek hooks skipped by id — they cannot bootstrap under GitHub-only worker egress; glassine hooks ran and passed)
- **Skipped**: code-simplifier polish pass — the fix diff is small and the byte-identical-generators invariant is exactly what mechanical simplification would break; external Codex/Gemini second opinions unavailable in this sandbox

### Commit

`fe4698de447f` — fix: resolve code review findings for PR #72 (all three inline threads replied to and resolved)

Verdict: approve

_[via gantry](https://gantry.hails.info/run/fabled-sharp-pebble)_

